PR #73, opened 2019-11-27
INFRA-2357 Add plugin site chart

timja:
Needs secrets

Tested on minikube, had a couple of issues with DNS / service routing.
Also tested on an AKS cluster and it seemed to work fine there

Edit:
minikube issue was:
https://github.com/kubernetes/minikube/issues/1568

Let me know what you think
cc @olblak

### DIFF
--- a/charts/plugin-site/templates/deployment.yaml
+++ b/charts/plugin-site/templates/deployment.yaml
@@ -20,6 +20,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+    {{- with .Values.hostAliases }}
+      hostAliases:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
--- a/charts/plugin-site/templates/deployment.yaml
+++ b/charts/plugin-site/templates/deployment.yaml
@@ -1,0 +1,77 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "plugin-site.fullname" . }}
+  labels:
+{{ include "plugin-site.labels" . | indent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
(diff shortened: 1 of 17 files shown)


oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/INFRA-2357

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/INFRA-2357 for it so that we can use the EPIC for all services

charts/plugin-site/templates/deployment.yaml
```diff
@@ -0,0 +1,81 @@
...
+{{ include "plugin-site.labels" . | indent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "plugin-site.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
```
olblak: @timja are you ok to add the maintainer label referencing your Github account?
This will be usefull when/if we create alert for specific services

olblak:
Can I also ask you to create a gpg key that you will use for this project and then update the [.sops](https://github.com/jenkins-infra/charts/blob/master/.sops.yaml#L13-L5) file, so you will be able to edit the secrets specific to the pluginsite just in case something wrong happen

olblak:
Just add an entry similar to jenkins-wik-exporter but without halkeye key of course

olblak:
And finally also add your account to the [CODEOWNER](https://github.com/jenkins-infra/charts/blob/master/CODEOWNERS) file, so you will be able to merge anything related to the pluginsite

config/default/plugin-site.yaml
```diff
@@ -0,0 +1,33 @@
...
+        - plugins.jenkins.io
+
+restApiUrl: https://plugins.jenkins.io/api
+
+replicaCount: 2
+
+resources:
+  limits:
+    cpu: 1500m
+    memory: 2048Mi
```
olblak: This value is definitely too high is more than probably too high and we'll adjust it once we have more infromation from the monitoring

timja: Sure, I didn't want to be too conservative on the limits, but needs monitoring data 👍

PR #75, opened 2019-11-28
INFRA-2359 Move javadoc to AKS

timja:


config/default/javadoc.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+replicaCount: 2
+
+resources:
+  limits:
+    cpu: 500m
+    memory: 1024Mi
+  requests:
+    cpu: 100m
+    memory: 256Mi
```
olblak: limits could use the same than request 100m and 256Mi

charts/javadoc/templates/secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.azureStorageAccountName }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: javadoc
+type: Opaque
+data:
+  azurestorageaccountname: {{ .Values.azureStorageAccountName | b64enc }}
+  azurestorageaccountkey: {{ .Values.azureStorageAccountKey | b64enc }}
```
olblak: Those two secrets should be documented in the readme and the default valules.yaml

PR #76, opened 2019-11-28
INFRA-2365 Add reports to AKS

timja:


charts/reports/values.yaml
```diff
@@ -0,0 +1,55 @@
+# Default values for plugins.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: nginx
+  tag: alpine
+  pullPolicy: IfNotPresent
```
olblak: As you are using the tag 'alpine', which is not a "fixed" image, can you set the pullPolicy to Always.
We don't take a lot of risks to always pull the latest available image, the nginx configuration remains quite standard and even if something is wrong, reports.jenkins.io is not critical and monitored

config/default/reports.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+replicaCount: 2
+
+resources:
+  limits:
+    cpu: 500m
+    memory: 1024Mi
+  requests:
+    cpu: 100m
+    memory: 256Mi
```
olblak: I am pretty confident that this container can use smaller values like:
``` 
 #   cpu: 100m
 #   memory: 128Mi
```

charts/reports/templates/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "reports.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "reports.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
```
olblak: Can you also specify a maintainer label ? here

charts/reports/templates/secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.azureStorageAccountName }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: reports
+type: Opaque
+data:
+  azurestorageaccountname: {{ .Values.azureStorageAccountName | b64enc }}
+  azurestorageaccountkey: {{ .Values.azureStorageAccountKey | b64enc }}
```
olblak: Those two secrets should be documented in the README, mentionned in the default values.yaml, and the .sops should be modified as well

PR #78, opened 2019-11-29
INFRA-2366 Jenkins admin chatbot to AKS

timja:


clusters/publick8s.yaml
```diff
@@ -18,3 +18,4 @@ helmfiles:
   - path: "../helmfile.d/oauth2-proxy.yaml"
   - path: "../helmfile.d/release-jenkins.yaml"
   - path: "../helmfile.d/plugin-site.yaml"
+  - path: "../helmfile.d/chatbot_jenkinsadmin.yaml"
```
timja: is this the right cluster? wasn't sure?

CODEOWNERS
```diff
@@ -4,5 +4,6 @@
 
 jenkins-wiki-exporter/ @halkeye
 plugin-site/ @timja
+chatbot_jenkinsadmin/ @timja
```
olblak:
```suggestion
chatbot_jenkinsadmin/ @timja @slide
```

olblak: @slide as he regularly contributes to this bot

charts/chatbot_jenkinsadmin/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for plugins.jenkins.io
+name: jenkinsadmin
+maintainers:
+  - name: timja
```
olblak:
```suggestion
  - name: timja
  - name: slide
```

charts/chatbot_jenkinsadmin/README.md
```diff
@@ -0,0 +1,28 @@
+# Plugin site
```
olblak:
```suggestion
# Chatbot Jenkinsadmin
```

olblak:
Before merging this PR, I would like to see if we can use github oauth token instead of github username/password, especially because the current user used has too many permission

timja:
> Before merging this PR, I would like to see if we can use github oauth token instead of github username/password, especially because the current user used has too many permission

I've never hit somewhere that I've not been able to use an oauth token on GitHub, it's required to work in general because a lot of orgs require 2fa to be enabled which prevents using a username and password

olblak:
> I've never hit somewhere that I've not been able to use an oauth token on GitHub, it's required to work in general because a lot of orgs require 2fa to be enabled which prevents using a username and password
Let rephrase this, how easy would it be to modify jenkins-infra/ircbot to use github token instead of username/password :D I created INFRA-2371

timja:
Shouldn't need any work at all, just use an oauth token instead of password

olblak:
let's try

PR #85, opened 2019-12-05
Rename Jenkinsfile to drop ci.jenkins.io picking it up

timja:
It will need org config adjusted on release.ci.jenkins.io, I can do that after merge.

This will stop the spam on every PR from the jenkinsadmin bot saying failure.

cc @halkeye 

Unless you have a better idea @olblak?


olblak:
I like this workaround for now, can you also update the jenkins configuration [here](https://github.com/jenkins-infra/charts/blob/master/config/default/jenkins.yaml#L67-L64) to use the right Jenkinsfile

timja:
> I like this workaround for now, can you also update the jenkins configuration [here](https://github.com/jenkins-infra/charts/blob/master/config/default/jenkins.yaml#L67-L64) [](/jenkins-infra/charts/blob/HEAD@{2019-12-05T16:19:30Z}/config/default/jenkins.yaml#L67-L64) to use the right Jenkinsfile

Done

timja:
Had to manually update the config, as the job that applies it wasn't working because it's jenkinsfile had to be removed.
would have needed to duplicate the file to have the automation transition it over

but didn't know there was automation for it when I moved the file 👅

PR #91, opened 2019-12-11
Enable CORS for plugin site

halkeye:
Need to be able to hit the API from front end, and easier than running my own fork.

### DIFF
--- a/config/default/plugin-site.yaml
+++ b/config/default/plugin-site.yaml
@@ -6,6 +6,10 @@ ingress:
     "nginx.ingress.kubernetes.io/rewrite-target": /$1
     "nginx.ingress.kubernetes.io/hsts": "true"
     "nginx.ingress.kubernetes.io/ssl-redirect": "true"
+    "nginx.ingress.kubernetes.io/enable-cors": "true"
+    "nginx.ingress.kubernetes.io/cors-allow-methods": "GET, OPTIONS"
+    "nginx.ingress.kubernetes.io/cors-allow-origin": "*"
+
 
   hosts:
     - host: plugins.jenkins.io


olblak:
I am not sure that understand what you are trying to solve here, is it just for your local dev environment?

halkeye:
For dev environment yes

Also if I succeed in making the plugin site static it shouldn't be a docker container but a cdn so different url

timja:
> For dev environment yes
> 
> Also if I succeed in making the plugin site static it shouldn't be a docker container but a cdn so different url

I assume it would still stay in a container, but be more like jenkins.io, backed by a storage account for the static content

For cost and ease of deployment?

oleg-nenashev:
So basically we expose the Plugin SITE API to the entire world? It should be okay, but I think it needs to be documented somewhere then

timja:
> So basically we expose the Plugin SITE API to the entire world? It should be okay, but I think it needs to be documented somewhere then

It's already exposed to the world (https://github.com/jenkins-infra/charts/blob/master/config/default/plugin-site.yaml#L13) , this just allows browsers not on the same domain to access it

olblak:
While I am not strongly opposed to this PR, I am not convinced that motivation here is worthwhile, and even if it's not harmful today who knows how this application will evolve? 
Couldn't you just run the docker image as explained [here](https://github.com/jenkins-infra/plugin-site-api/#run-docker-plugin-site-api) where the data url is fetched from [ci.jenkins.io](https://ci.jenkins.io/job/Infra/job/plugin-site-api/job/generate-data/)

olblak:
@halkeye @timja @oleg-nenashev what are you thought about my last comment?
As I said I am not strongly opposed to this PR, so I am also fine to merge this to move forward

timja:
> @halkeye @timja @oleg-nenashev what are you thought about my last comment?
> As I said I am not strongly opposed to this PR, so I am also fine to merge this to move forward

I can't really see any harm in exposing it, other than potentially allowing other websites to request the API from the browser side.

It's quite rare to have to change the plugin-site-api, more work is being done on the plugin-site itself so I see why @halkeye has made this PR

halkeye:
<_< ah merged

I was going to say I figured out a solution for dev, but yay! makes things easier!

PR #96, opened 2019-12-17
Fix maintainer label

timja:


timja:
> You cheat, it would have been nicer to have every maintainer in the label something like "maintainer1_ maintainer2"
> But I can't find an easy and clean way to do it with helm so let move this forward
> Thanks for this PR

neither, this was hard enough 😉

halkeye:
https://stackoverflow.com/questions/47668793/helm-generate-comma-separated-list

I've used this before which works really good

https://github.com/halkeye-helm-charts/powerdns-recursor/blob/master/templates/_helpers.tpl#L59-L62

Then you either `{{- include "powerdns-recursor.joinListWithComma" .Values.zones | quote}}` or `"{{- include "powerdns-recursor.joinListWithComma" .Values.zones }}"` with quotes

See https://github.com/halkeye-helm-charts/powerdns-recursor/blob/6c645dab927b426ba8e5a9859be1375762631450/templates/deployment.yaml#L34

timja:
> [stackoverflow.com/questions/47668793/helm-generate-comma-separated-list](https://stackoverflow.com/questions/47668793/helm-generate-comma-separated-list)
> 
> I've used this before which works really good
> 
> [halkeye-helm-charts/powerdns-recursor:templates/_helpers.tpl@`master`#L59-L62](https://github.com/halkeye-helm-charts/powerdns-recursor/blob/master/templates/_helpers.tpl#L59-L62) [](/halkeye-helm-charts/powerdns-recursor/blob/HEAD@{2019-12-17T18:28:21Z}/templates/_helpers.tpl#L59-L62)
> 
> Then you either `{{- include "powerdns-recursor.joinListWithComma" .Values.zones | quote}}` or `"{{- include "powerdns-recursor.joinListWithComma" .Values.zones }}"` with quotes
> 
> See [halkeye-helm-charts/powerdns-recursor:templates/deployment.yaml@`6c645da`#L34](https://github.com/halkeye-helm-charts/powerdns-recursor/blob/6c645dab927b426ba8e5a9859be1375762631450/templates/deployment.yaml#L34)

That would work but you would still be separating with underscores which is quite weird

PR #97, opened 2019-12-17
Remove comment breaking build

timja:


timja:
> Is it worth converting to a csv? It wouldn't be queryable, but would help know who to contact

Given the number of people who have access I'm not hugely sure the worth of having the label =/

CODEOWNERS / github show it just as well.

But @olblak seemed keen on it

olblak:
> Is it worth converting to a csv? It wouldn't be queryable, but would help know who to contact
Nope because label format is really restrictive and doesn't accept ',' or ';' [doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

> But @olblak seemed keen on it
The idea is to simplify our alerting rules, by having pull-based configuration instead of a push based approach.

The simplest approach would be to specify a Github team instead of a GitHub account.
Or we just don't bother about that until we rework the monitoring

PR #98, opened 2019-12-18
Remove HSTS on jenkins.io

timja:
Breaks accessing mirrors in browser
See https://github.com/kubernetes/ingress-nginx/issues/549


timja:
> I didn't catch the hsts set to true

pretty hard to catch that from looking at the code, the annotation from the ACS cluster doesn't work on this ingress controller.

PR #101, opened 2020-01-07
Update plugins helm chart to pull files from infra

halkeye:


charts/plugin-site/values.yaml
```diff
@@ -2,12 +2,46 @@
...
+  resources:
+    limits:
+      cpu: 2000m
+      memory: 2048Mi
+    requests:
+      cpu: 1000m
+      memory: 1024Mi
+  #resources:
+  #  limits:
+  #    cpu: 100m
```
timja: I assume these are the resources of you are using new mode?

charts/plugin-site/templates/deployment.yaml
```diff
@@ -22,9 +22,79 @@ spec:
...
+    {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+---
+apiVersion: apps/v1
```
timja: I think it would be cleaner if each deployment was in its own file, more obvious where to look to find the resource to me

charts/plugin-site/values.yaml
```diff
@@ -17,43 +51,36 @@ service:
   type: ClusterIP
 
 # Please define ingress settings into environment variable
-#ingress:
-#  enabled: true
+ingress:
+  enabled: disabled
```
timja: false rather than disabled? It’s checking for a Boolean but if it’s not true it’s ignored

halkeye: oops

I mean honestly its never going to be false, but yea, i'll fix

charts/plugin-site/values.yaml
halkeye: yea, and honestly probably a lot smaller than that, I don't fully understand the numbers :)

charts/plugin-site/values.yaml
timja: Only reason would be someone accessing it on NodePort because they can’t get ingress working locally I think, and even then it causes no harm not to have it enables

timja: How about just enabling it?

charts/plugin-site/values.yaml
timja: This is pretty small I wouldn’t go smaller and should be fine for a static site

halkeye:
```
$ kubectl top pod
NAME                                                      CPU(cores)   MEMORY(bytes)
jenkins-plugins-io-plugin-site-backend-6d7c96d98b-xvb4d   7m           489Mi
jenkins-plugins-io-plugin-site-frontend-9f6665996-ng799   1m           3Mi
```

charts/plugin-site/values.yaml
```diff
@@ -2,12 +2,46 @@
...
+    #tag: 44-14cfe6
+    pullPolicy: IfNotPresent
+  port: 80
+  resources:
+    limits:
+      cpu: 2000m
+      memory: 2048Mi
+    requests:
+      cpu: 1000m
+      memory: 1024Mi
```
olblak: The frontend doesn't probably need that amount of memory and cpu.

halkeye: the javascript version does, I had it set locally to 100m cpu, and it took 15 min to test, compile, and startup.
at 1000m, it took a couple seconds

The old image is really not optimized in any way.

timja: Just need to remember to update the resources when we enable the new static site

halkeye:
Tested with flexvolume as I don't have azure:
```
backend:
  replicaCount: 1
frontend:
  replicaCount: 1

htmlVolume:
  flexVolume:
    driver: juliohm/cifs
    options:
      opts: guest,uid=1000,gid=1000
      server: odin
      share: /gatsby-plugin-site
```

halkeye:
> Doesn’t the readme need an update on running locally now that you need static files in place?

THERES A README??!?!?!?

Um, probably

I'm not sure how you'd run it locally. ~Does minikube have any sort of local disk mounter?~

Should I just document hostpath?

https://kubernetes.io/docs/concepts/storage/volumes/#example-pod-2

halkeye:
okay, https://kubernetes.io/docs/concepts/storage/volumes/#hostpath is a better link

Documented

timja:
Take a look at the other ones that require static files like Jenkins io they are a bit more detailed I think

timja:
Minikube does have a command to mount a folder to the node, it’s documented in the readme’s that require it

timja:
e.g. 
https://github.com/jenkins-infra/charts/tree/master/charts/javadoc#running-this-yourself

config/default/plugin-site.yaml
```diff
@@ -26,12 +20,11 @@ ingress:
...
-  limits:
-    cpu: 2000m
-    memory: 2048Mi
-  requests:
-    cpu: 1000m
-    memory: 1024Mi
+htmlVolume:
+  azureFile:
+    secretName: plugin-site
+    shareName: prodpluginsite
```
olblak: shareName 'pluginsite' as defined [here](https://github.com/jenkins-infra/azure/blob/936993e6306247920552c84d659106a93ca0d01d/plans/pluginsite.tf#L28)

timja:
```suggestion
    shareName: pluginsite
```

halkeye: sooooo, i don't think thats true actually.

https://github.com/jenkins-infra/azure/blob/936993e6306247920552c84d659106a93ca0d01d/plans/pluginsite.tf#L6

21:34 <rtyler> our Azure resources are prefixed with prod fwiw

https://github.com/jenkins-infra/plugin-site/blob/master/Jenkinsfile#L74-L84

timja: the resource group is prefixed but the share is hardcoded

halkeye: okay, the naming scheme for azure seems different with each tool, so i'll defer

olblak: ShareName is one level below than the azure file storage.
shareName name needs to be unique per azure file storage, where azure file storage need to be unique.

olblak:
I just updated plugin-site credentials and now you should be able to modify that secret file, even if in the current state you can't have access to the azure account

PR #103, opened 2020-01-08
[Infra-2364] Deploy ldap on publick8s

olblak:
This PR is part of [INFRA-1797](https://issues.jenkins-ci.org/browse/INFRA-1797)


charts/ldap/templates/persistentVolumeClaim.yaml
```diff
@@ -0,0 +1,27 @@
...
+    requests:
+      storage: 100Gi
+  selector:
+    matchLabels:
+      role: ldap-backup
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "ldap.fullname" . }}-data
```
timja: I can't find the PV for this in the PR? is it defined somewhere else?

olblak: This is the default behavior, Kubernetes provisions a new azure disk storage and then attach it to the pod, which means that the lifecycle is tied to kubernetes cluster.

image

More information on this application design [here](https://github.com/jenkins-infra/iep/tree/master/iep-008#data)

olblak:
> README for the chart?

Good point

charts/ldap/templates/persistentVolumeClaim.yaml
timja: Ah, I see. thanks for the link, didn't realise the backup/restore was so fast

charts/ldap/README.md
```diff
@@ -0,0 +1,21 @@
...
+- [Docker](https://github.com/jenkins-infra/ldap)
+- [DockerHub](https://hub.docker.com/r/jenkinsciinfra/ldap)
+- [IEP](https://github.com/jenkins-infra/iep/tree/master/iep-008)
+
+
+## Running this yourself
+
+`helmfile -f helmfile.d/ldap.yaml apply`
+
+ ### Minikube gotchas
```
timja: extra space was there:
```suggestion
### Minikube gotchas
```

olblak: thx

PR #108, opened 2020-01-14
[plugin-site] rename env variable to GITHUB_SECRET to match code

halkeye:
as far as I can tell from https://github.com/jenkins-infra/plugin-site-api/blob/master/src/main/java/io/jenkins/plugins/services/impl/GithubExtractor.java#L40 the env variable should be GITHUB_SECRET


halkeye:
SCARY MERGE

PR #112, opened 2020-01-18
From enable-cors.org, enable cors for site manifest

halkeye:


halkeye:
Ack sorry for no description. Too many PRs created.


No. Since I merged the other one this is safe to close.

PR #113, opened 2020-01-20
Add a chart that setups a proxy from new cluster to old

halkeye:
mostly to make it easier to handle https


halkeye:
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-certificate-authentication if needed

olblak:
This PR doesn't make sense as neither confluence or Jira are running on Kubernetes. 
They are both run machines sponsored by OSUOSL, and managed by our puppet repository "[jenkins-infra/jenkins-infra|https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/manifests/confluence.pp]"

halkeye:
Yea kill it. I thought it would be easier to let nginx-ingress handle certs instead of manually having to restart each time

PR #119, opened 2020-01-27
Increase wiki exporter CPU limits

timja:
When exporting heavy wiki pages or lots in a short time the current CPU limit is not enough


timja:
I can't merge as @halkeye is CODEOWNER and hopefully he is sleeping 😄

PR #122, opened 2020-01-28
Deploy Mirrorbits

olblak:
This Pull request deploys [mirrorbits](https://github.com/etix/mirrorbits) and a Redis database on Kubernetes.

In order to consider this PR as a full replacement of mirrors.jenkins.io, we still need to validate the following things
- [x] Validate that it's effectively working :) 
- [ ] Automate mirror configuration
- [ ] Make it working for Redis sentinel

Currently, this service is accessible from the VPN but empty at the moment

Update /etc/hosts with [release.mirrors.jenkins.io](https://release.mirrors.jenkins.io/?mirrorstats ) set to `10.0.2.5`

Example: https://release.mirrors.jenkins.io/debian/jenkins_2.218_all.deb?mirrorlist

or https://release.mirrors.jenkins.io/debian/jenkins_2.218_all.deb?mirrorstats
https://release.mirrors.jenkins.io/debian/jenkins_2.218_all.deb?stats

### DIFF
--- a/charts/mirrorbits/templates/secret.yaml
+++ b/charts/mirrorbits/templates/secret.yaml
@@ -1,0 +1,18 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "mirrorbits.fullname" . }}-binary
+type: Opaque
+data:
+  azurestorageaccountname: {{ .Values.storageAccountName | b64enc }}
+  azurestorageaccountkey: {{ .Values.storageAccountKey | b64enc }}
+
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "mirrorbits.fullname" . }}
+type: Opaque
+data:
+  mirrorbits.conf: {{ .Values.mirrorbits.conf | b64enc }}
--- a/charts/mirrorbits/templates/secret.yaml
+++ b/charts/mirrorbits/templates/secret.yaml
@@ -2,17 +2,20 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "mirrorbits.fullname" . }}-binary
-type: Opaque
-data:
-  azurestorageaccountname: {{ .Values.storageAccountName | b64enc }}
-  azurestorageaccountkey: {{ .Values.storageAccountKey | b64enc }}
-
----
-apiVersion: v1
-kind: Secret
-metadata:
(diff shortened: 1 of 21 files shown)


charts/mirrorbits/templates/secret.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "mirrorbits.fullname" . }}
+type: Opaque
+data:
+  mirrorbits.conf: {{ .Values.mirrorbits.conf | b64enc }}
```
timja: why does this need to be a secret and not a configmap?

olblak: because you have redis password in the configuration

olblak:
I did a few more tests, and made it working with few limitations:
1) mirror configuration needs to be manually done as Mirrorbits doesn't provide a different way than using the cli
2) I wasn't able to use redis sentinelle for High availibilty purpose.

olblak:
Mirrorbit requires a folder with all the files that can be provided.
If for some reason the requested file has a hash that mismatched between one of the mirrors and Mirrorbits then the mirror is not used.

Similar to [here](https://github.com/jenkins-infra/charts/pull/130#issue-378916917), they are different ways to synchronize data on Mirrorbit:

1) Mirrobits data is populated on a push approach like from [here](https://github.com/jenkins-infra/jenkins-infra/blob/fd567f7c295a700fac5ea30fa4e5f8cc675487d1/dist/profile/files/mirrorbrain/sync.sh#L48). this brings strong dependency on sync.sh and Mirrrobits becomes the source of truth for artifacts
2) Mirrorbits data is generated on a pull approach, similar to [here|https://github.com/jenkins-infra/charts/pull/130#issue-378916917], in this case, we need to ensure that Mirrorbits is always in sync with the source of truth.

I have a preference for the latter approach where we consider Mirroribts as an independent service we still have the capability to manually trigger a sync using `kubectl exec...`

olblak:
I just opened a Pull request [here](https://github.com/jenkins-infra/azure/pull/144) to provision an azure file storage to hold those files

olblak:
I am going to merge this pull request, I manually deployed get.jenkins.io using this code, and I am not planning to work on it for now.

PR #130, opened 2020-02-24
[INFRA-2466] Deploy archives.jenkins-ci.org on publick8s

olblak:
This pull request defines a Helm chart that can be used to deploy a HTTP/HTTPS mirror using RSYNC for synchronization.
The purpose of it is to move http://archives.jenkins-ci.org from Rackspace on our k8s cluster and also providing HTTPS.

Remarks:

1) Instead of pushing updates from pkg.jenkins.io to archives.jenkins-ci.org as defined [here](https://github.com/jenkins-infra/jenkins-infra/blob/fd567f7c295a700fac5ea30fa4e5f8cc675487d1/dist/profile/files/mirrorbrain/sync.sh#L48), this approach pulls data few times per hours from a public repository and so makes it totally independent of the release process or the Jenkins infrastructure itself.
Anybody can install this chart and then ask to add the new mirror to mirrors.jenkins.io

2) The mirror deployed on publick8s uses azure disk volume as it's the default storage class AKS, so the lifecycle of data is tighted to the lifecycle of publick8s.

### DIFF
--- a/charts/mirror/templates/configmap.yaml
+++ b/charts/mirror/templates/configmap.yaml
@@ -241,7 +241,7 @@ data:
         # It can be "All", "None", or any combination of the keywords:
         #   AllowOverride FileInfo AuthConfig Limit
         #
-        AllowOverride None
+        AllowOverride FileInfo
     
         #
         # Controls who can get stuff from this server.
--- a/charts/mirror/templates/configmap.yaml
+++ b/charts/mirror/templates/configmap.yaml
@@ -1,29 +1,517 @@
+# Disable  configmap for now
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "mirror.fullname" . }}
 data:
-  default.conf: |
-    server {
-      listen       80;
-      server_name  localhost;
-  
-      location / {
-          autoindex on;
-          root   /usr/share/nginx/html;
-          index  index.html index.htm;
-      }
+  httpd.conf: |
+    ServerRoot "/usr/local/apache2"
+    
+    Listen 80
+    
+    #
+    # Dynamic Shared Object (DSO) Support
+    #
+    # To be able to use the functionality of a module which was built as a DSO you
+    # have to place corresponding `LoadModule' lines at this location so the
+    # directives contained in it are actually available _before_ they are used.
+    # Statically compiled modules (those listed by `httpd -l') do not need
+    # to be loaded here.
+    #
+    # Example:
+    # LoadModule foo_module modules/mod_foo.so
+    #
+    LoadModule mpm_event_module modules/mod_mpm_event.so
+    #LoadModule mpm_prefork_module modules/mod_mpm_prefork.so
+    #LoadModule mpm_worker_module modules/mod_mpm_worker.so
+    LoadModule authn_file_module modules/mod_authn_file.so
+    #LoadModule authn_dbm_module modules/mod_authn_dbm.so
+    #LoadModule authn_anon_module modules/mod_authn_anon.so
+    #LoadModule authn_dbd_module modules/mod_authn_dbd.so
+    #LoadModule authn_socache_module modules/mod_authn_socache.so
+    LoadModule authn_core_module modules/mod_authn_core.so
+    LoadModule authz_host_module modules/mod_authz_host.so
+    LoadModule authz_groupfile_module modules/mod_authz_groupfile.so
+    LoadModule authz_user_module modules/mod_authz_user.so
+    #LoadModule authz_dbm_module modules/mod_authz_dbm.so
+    #LoadModule authz_owner_module modules/mod_authz_owner.so
+    #LoadModule authz_dbd_module modules/mod_authz_dbd.so
+    LoadModule authz_core_module modules/mod_authz_core.so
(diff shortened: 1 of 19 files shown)


charts/mirror/templates/configmap.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "mirror.fullname" . }}
+data:
+  default.conf: |
+    server {
+      listen       80;
+      server_name  localhost;
```
timja: is this right?

olblak: technically yes, it works because it's the default and only server configuration, we use it everywhere in our infra

olblak:
So in the current state, this chart cannot be used as a replacement for archives.jenkins-ci.org. "ftp-osl.osuosl.org/jenkins/", the source of truth here, only contains one-year-old packages as defined [here](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/files/mirrorbrain/sync.sh#L21L20)
Currently only archives.jenkins-ci.org contains every packages.

charts/mirror/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+            tcpSocket:
+             port: 80
+          readinessProbe:
+            httpGet:
+              path: /
+              port: http
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+          {{- if $.Values.persistent.volumeClaim.enabled }}
```
halkeye: what does the $ do here? it should be just .Values right?

daniel-beck:
> Instead of pushing updates from pkg.jenkins.io to archives.jenkins-ci.org as defined here, this approach pulls data few times per hours from a public repository and so makes it totally independent of the release process or the Jenkins infrastructure itself.

How can I ensure that artifacts are quickly made available on mirrors? `sync.sh` is the first part of a two step process to ensure downloads are available before I announce security fixes.

(@olblak See https://github.com/jenkins-infra/runbooks/tree/master/security#release-from-staging steps 6 through 9)

olblak:
> > Instead of pushing updates from pkg.jenkins.io to archives.jenkins-ci.org as defined here, this approach pulls data few times per hours from a public repository and so makes it totally independent of the release process or the Jenkins infrastructure itself.
> 
> How can I ensure that artifacts are quickly made available on at least one mirror? `sync.sh` is the first part of a two step process to ensure downloads are available before I announce security fixes.
> 
> (@olblak See https://github.com/jenkins-infra/runbooks/tree/master/security#release-from-staging steps 6 through 9)

Considering that archives.jenkins-ci.org contains every artifacts generated, more than mirrors, it has to be the source of truth and therefore cannot pull it's data from a third mirror.
We need to create a new storage account for archives and push new artifacts as soon as they are generated

olblak:
I just opened a Pull request [here](https://github.com/jenkins-infra/azure/pull/144) to provision an azure file storage to hold those files

charts/mirror/templates/deployment.yaml
olblak: I though it was needed but apparently not as documented [here](https://helm.sh/docs/chart_template_guide/control_structures/)

timja:
conflicts

olblak:
This pull request will deploy two services
* mirror.azure.jenkins.io which synchronize its files with `rsync://ftp-osl.osuosl.org/jenkins/`, 
considering that we heavily rely on htaccess files, I may switch the default nginx to apache

* archives.jenkins-ci.org which is mount files from an azure file storage populated from [here](https://github.com/jenkins-infra/jenkins-infra/pull/1446)

config/default/archives.yaml
```diff
@@ -0,0 +1,58 @@
...
+synchronize:
+  enabled: false
+
+persistent:
+  volumeClaim:
+    name: archives-binary
+    enabled: true
+    spec:
+      accessModes:
+        - ReadWriteMany
```
timja: does it need to write? isn't it just read?

olblak: You are right it needs to be read-only

olblak:
An example of a htacess rule.
This url https://archives.jenkins.io/osx-rc/latest should be redirected to https://archives.jenkins.io/osx-rc/jenkins-2.0.pkg

config/default/archives.yaml
olblak: Fixed

olblak:
So to recap this pull request is now ready to manage two services 
* mirror.azure.jenkins.io, pull based, it works standalone and fetch its data from `rsync://ftp-osl.osuosl.org/jenkins/` every 5minutes
* archives.jenkins.io, push based, data need to be push from pkg.jenkins.io and it has the same files storage than get.jenkins.io, currently uploading file is disabled.

olblak:
Considering that archives.jenkins-ci.org don't have to move anymore, I propose to keep the work being done here but just using a different url archives.azure.jenkins.io

And we'll also add https support to https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/manifests/archives.pp

PR #134, opened 2020-02-25
Lock grafana

timja:


helmfile.d/grafana.yaml
```diff
@@ -9,3 +9,5 @@ releases:
         - "../config/default/grafana.yaml"
       secrets:
         - "../secrets/config/grafana/secrets.yaml"
+bases:
+  - 00-repositories.yaml
```
halkeye: What's that file?

timja: the repos, apparently the default parent implied way is deprecated, without it, it wouldn't generate the lock file, (i manually edited the file so downgrade, hopefully it works)

halkeye:
https://github.com/dependabot/feedback/issues/174 :( one day one day

timja:
renovate has it (an alternative to dependabot)

PR #136, opened 2020-02-26
Revert "Lock grafana (#134)"

olblak:
This reverts commit 09d0e49b0db95141a16d97bc33a5cc9185f9b78c.

Grafana is not a critical service for us so I prefer to always install the latest version and fix issues when they occur than locking to  a specific version and then taking the risk to run outdated version.

### DIFF
--- a/helmfile.d/grafana.yaml
+++ b/helmfile.d/grafana.yaml
@@ -9,5 +9,3 @@ releases:
         - "../config/default/grafana.yaml"
       secrets:
         - "../secrets/config/grafana/secrets.yaml"
-bases:
-  - 00-repositories.yaml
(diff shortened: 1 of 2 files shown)


olblak:
As explained  [here](https://github.com/helm/charts/tree/master/stable/grafana#to-500), I temporary added `force: true` to 'helmfile.d/grafana.yaml' then manually run helmfile -f helmfile.d/grafana.yaml apply. 

I didn't commit this change as it was a temporary workaround

helmfile.d/grafana.yaml
```diff
@@ -9,5 +9,3 @@ releases:
         - "../config/default/grafana.yaml"
       secrets:
         - "../secrets/config/grafana/secrets.yaml"
-bases:
```
timja: this bit shouldn't really have been reverted, parents inheriting by default is deprecated:
https://github.com/roboll/helmfile/issues/688

PR #139, opened 2020-03-04
Various Jenkins modification

olblak:
Sorry for this big PR but basically I am proposing following changes

- Bump Jenkins to latest version by default
- Switch from kubernetes-credential-provider to jcasc by default for credentials, so the env-jenkins-release chart only configures specific volume for the release environment
- Deploy a new Jenkins service called infra.ci.jenkins.io where we run k8smgmnt job instead of running that job on release.ci so release.ci.jenkins.io access can be more restrictive
- Use an experimental tool that I wrote to automate application update more information [here](https://github.com/olblak/updatecli/), a pull request example is [here](https://github.com/olblak/charts/pull/11/files), as a first test, I would like to use it for jenkins-wiki-exporter and pluginsite-api

Remark: I didn't commit the secrets on chart-secrets


config/default/jenkins-infra.yaml
```diff
@@ -0,0 +1,127 @@
...
+                        - string:
+                            scope: GLOBAL
+                            id: "updatecli-github-token"
+                            secret: '${UPDATECLI_GITHUB_TOKEN}'
+                            description: Github Token used by updatecli to update version
+                        - basicSSHUserPrivateKey:
+                            scope: GLOBAL
+                            id: "charts-secrets"
+                            username: ${SSH_CHARTS_SECRETS_USERNAME}
+                            #passphrase: ${SSH_CHARTS_SECRETS_PASSWORD}
```
timja: ?

config/default/jenkins-infra.yaml
```diff
@@ -0,0 +1,127 @@
...
+                            description: "SSH privkey used to access jenkins-infra/charts-secret"
```
timja:
```suggestion
                            description: "SSH privkey used to access jenkins-infra/charts-secrets"
```

config/default/jenkins-infra.yaml
```diff
@@ -0,0 +1,127 @@
...
+                          userSearch: cn={0}
+                      cache:
+                        size: 100
+                        ttl: 300
+              matrix-settings: |
+                jenkins:
+                  authorizationStrategy:
+                    globalMatrix:
+                      grantedPermissions:
+                        - "Overall/Administer:admins"
```
timja: what about?
```suggestion
                        - "Overall/SystemRead:all"
```

config/default/jenkins-infra.yaml
```diff
@@ -0,0 +1,127 @@
...
+                              spec:
+                                tolerations:
+                                - key: "os"
+                                  operator: "Equal"
+                                  value: "windows"
+                                  effect: "NoSchedule"
+              jobs-settings: |
+                jobs:
+                  - script: >
+                      multibranchPipelineJob('k8smgmt') {
```
timja: do you plan to move some of the other infra jobs here as well? like azure? and plugin-site-api?

halkeye:
I totally forgot to review this today, but doesn't the change in secrets to env variables, expose those env variables inside of jenkins master?

timja:
> I totally forgot to review this today, but doesn't the change in secrets to env variables, expose those env variables inside of jenkins master?

they aren't env variables, 
https://github.com/jenkins-infra/charts/pull/139/files#diff-20f45507eb92beac86005a6fbc038ba4R9-R11

They are a k8s secret which is mounted on the master file system,
pretty much the same as before

config/default/jenkins-infra.yaml
```diff
@@ -0,0 +1,127 @@
...
+                            containers:
+                              - name: jnlp
+                                image: "jenkins4eval/jnlp-agent:latest-windows"
+                                resourceLimitCpu: "500m"
+                                resourceLimitMemory: "512Mi"
+                                resourceRequestCpu: "500m"
+                                resourceRequestMemory: "512Mi"
+                                args: "^${computer.jnlpmac} ^${computer.name}"
+                                alwaysPullImage: true
+                            label: "jnlp-windows"
```
halkeye: Shouldn't the label also be on the Linux one?

PodTemplates.yaml
```diff
@@ -27,6 +27,26 @@ spec:
     securityContext:
       privileged: false
     tty: true
+  - args:
+    - "99d"
+    command:
```
halkeye: Is the sleep a left behind from debugging?

helmfile.d/jenkins-infra.yaml
```diff
@@ -0,0 +1,17 @@
...
+releases:
+    - name: default-jenkins-infra
+      chart: ../charts/jenkins
+      namespace: jenkins-infra
+      values:
+        - "../config/default/jenkins-infra.yaml"
+      secrets:
+        - "../secrets/config/jenkins-infra/jenkins.yaml"
+      set:
+        - name: namespace
```
halkeye: I dont think you need namespace above and here

updateCli.d/pluginsite.yaml
```diff
@@ -0,0 +1,49 @@
...
+    repository: "plugin-site-api"
+    token: ""
+    username: "olblak"
+    version: "latest"
+conditions:
+  docker:
+    name: "Docker Image Published on Registry"
+    kind: dockerImage
+    spec:
+      image: "jenkinsciinfra/plugin-site"
```
halkeye: Shouldn't this be the backend image, so plugin-site-api?

config/default/jenkins-infra.yaml
olblak: I removed passphrase as it's not needed

olblak: I removed passphrase as it's not needed

config/default/jenkins-infra.yaml
olblak: Would it be safe to enabled the systemread by default?

config/default/jenkins-infra.yaml
olblak: Nice catch, this label is not needed.

PodTemplates.yaml
olblak: It's just a way to be sure that the container will be running for the lifetime of the job.
The main container process run 'sleep 99d' and the the jenkinsfile can run as many sh'' steps, this has to be done for every container in a pod

olblak: Jenkins is then responsible to cleanup the pods once done

helmfile.d/jenkins-infra.yaml
olblak: Yes I do need to configure that namespace to tell to the k8s plugin where to spinup pods, I just need to update it to the right place, thanks

updateCli.d/pluginsite.yaml
olblak: Yes it should even if for now, it won't work because, jenkinsciinfra/plugin-site-api doesn't have a tag similar to github release like jenkins-wiki-exporter :)

config/default/jenkins-infra.yaml
olblak: I don't think so because that this service needs cluster-admin permission to manage the cluster, I want to keep it only the vpn and I wouldn't allow to many people with access on it.

config/default/jenkins.yaml
```diff
@@ -124,10 +206,7 @@ jenkins:
                     globalMatrix:
                       grantedPermissions:
                         - "Overall/Administer:release-core"
-                        - "Overall/Read:release-core-dev"
-                        - "Job/Build:release-core-dev"
-                        - "Job/Cancel:release-core-dev"
-                        - "Job/Read:release-core-dev"
+                        - "Overall/SystemRead:all"
```
timja: Because of weirdness in permissions and how they applied you also need these:
```suggestion
                        - "Overall/SystemRead:all"
                        - "Overall/Read:all"
                        - "Job/Read:all"
```
Administer is the only god permission, everything else needs applying explicitly

I'm planning to try get a 'full read' role added in role-strategy-plugin which would make this simpler as you don't need to care about specific permissions it just grants all the read ones that make sense

PR #168, opened 2020-03-31
Release environment must use jenkins-infra repository

olblak:
* Switch release.ci.jenkins.io to lts version and automate its upgrade when a new docker image is available based on digest
* Rename jenkins.yaml configuration file to jenkins-release.yaml to be more explicit on the content
* Fix release.ci.jenkins.io packaging job to use jenkins-infra/release instead of olblak/release

### DIFF
--- a/config/default/jenkins-release.yaml
+++ b/config/default/jenkins-release.yaml
@@ -2,7 +2,7 @@ clusterAdminEnabled: false
 jenkins:
     master:
         image: jenkins/jenkins@sha256
-        imageTag: af0110d8f70d2fbaf9429c321dc4e3bc5aae12f708c655f4018e7ed4be0974f5
+        imageTag: 973a0d1dd45f0706ac720c50f14f8ea7de6a80583cc0100af76d4288a7d89714
         JCasC:
             enabled: true
             configScripts:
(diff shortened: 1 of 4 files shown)


updateCli.d/jenkins-release.yaml
```diff
@@ -0,0 +1,21 @@
+source:
+  kind: dockerDigest
+  spec:
+    image: "jenkins/jenkins"
+    tag: "lts"
```
timja, line 5: should this be?:
```suggestion
    tag: "lts-jdk11"
```

olblak: Yes, nice catch then we also have to update the digest

config/default/jenkins-release.yaml
```diff
@@ -1,6 +1,8 @@
 clusterAdminEnabled: false
 jenkins:
     master:
+        image: jenkins/jenkins@sha256
+        imageTag: af0110d8f70d2fbaf9429c321dc4e3bc5aae12f708c655f4018e7ed4be0974f5
```
olblak, line 5:
```suggestion
        imageTag: 973a0d1dd45f0706ac720c50f14f8ea7de6a80583cc0100af76d4288a7d89714
```

olblak: this digest correspond to the latest jenkins/jenkins:lts-jdk11

PR #169, opened 2020-04-01
Install the powershell plugin

slide:
This will make some of the scripting easier for release

### DIFF
--- a/charts/jenkins/values.yaml
+++ b/charts/jenkins/values.yaml
@@ -79,6 +79,7 @@ jenkins:
     - pipeline-stage-tags-metadata
     - pipeline-stage-view
     - plain-credentials
+    - powershell
     - prometheus
     - scm-api
     - script-security


charts/jenkins/values.yaml
```diff
@@ -79,6 +79,7 @@ jenkins:
     - pipeline-stage-tags-metadata
     - pipeline-stage-view
     - plain-credentials
+    - powershell
```
timja, line 82: is it needed? the docs say it doesn't integrate with pipeline? https://plugins.jenkins.io/powershell/

oleg-nenashev: Powershell is supported negatively in the Pipeline Durable Task plugin: https://jenkins.io/blog/2017/07/26/powershell-pipeline/ . So I am not sure why we need this plugin TBH. @slide ?

slide: You are correct.

PR #170, opened 2020-04-03
Release env, add ssh key to upload artifacts on pkj.j.io

olblak:
A new ssh key used to push artifacts generated by the release environment to pkg.jenkins.io

### DIFF
--- a/config/default/jenkins-release.yaml
+++ b/config/default/jenkins-release.yaml
@@ -93,6 +93,15 @@ jenkins:
                                 privateKey: ${SSH_RELEASE_PRIVKEY}
                         - basicSSHUserPrivateKey:
                             scope: GLOBAL
+                            id: "pkgserver"
+                            username: ${SSH_PKGSERVER_USERNAME}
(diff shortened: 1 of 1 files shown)


config/default/jenkins-release.yaml
```diff
@@ -91,6 +91,15 @@ jenkins:
                             privateKeySource:
                               directEntry:
                                 privateKey: ${SSH_RELEASE_PRIVKEY}
+                        - basicSSHUserPrivateKey:
+                            scope: GLOBAL
+                            id: "pkgserver"
+                            username: ${SSH_PKGSERVER_USERNAME}
```
timja, line 97: secrets not added yet?
https://github.com/jenkins-infra/charts-secrets/blob/master/config/release/jenkins/secrets.yaml#L1

olblak: Better when I run git push, currently it's a none working ssh key, as  we don't want to accidently push any artifacts to pkg.jenkins.io, and I'll more than probably setup a ssh bastion to test the scripts

PR #173, opened 2020-04-08
Add Job/Read access to infra ci

timja:


timja:
> Should not be a problem after `Overall/SystemRead:all` above, right?

its intended that everyone has read access, but Olivier didn't know about SystemRead not granting other permissions, and i missed when i reviewed

PR #185, opened 2020-04-20
Update to build with for for INFRA-2545

slide:
This has a fix for the hosting process (see issues.jenkins-ci.org/browse/INFRA-2545)


halkeye:
@olblak we need an issue to make these semversioned releases too, so your update cli can auto do them.

PR #198, opened 2020-04-30
Switch ldap certificate to letsencrypt

olblak:
This pull request has already been tested and deployed, I took two shortcuts here
* The ldap container won't detect a certificate change so this is something that need to implemented
* The secret generated by the certificate doesn't add any value to ca.crt, so I did it manually


olblak:
I am going to merge this, but I'll also have to work on https://issues.jenkins-ci.org/browse/INFRA-2600

PR #201, opened 2020-05-04
Automate ldap docker image tag update

olblak:


olblak:
I am going to wait that the release is done before playing with ldap :p

PR #205, opened 2020-05-05
Fix ldap cron image tag

olblak:
1. Fix docker image tag used for the cron daemon
2. Automate both slapd and crond docker image tag by using digest


olblak:
I am moving forward as currently ldap.jenkins.io is down because the crond container uses an invalid digest 

```
 Failed to apply default image tag "jenkinsciinfra/ldap@sha256:cron-6a272bb54780919010eebee33fc2cafb8406ca42af0209c7b6b8ed2f1e38a3b6"
```

PR #206, opened 2020-05-13
Bump updatecli version && Update mirrorbits doc

olblak:


olblak:
> I love this. It went from complaining and wanting to run a local copy to not deal with mess, to a public mirror provider. It just makes me feel warm

If you are talking about mirrorbits, kind reminder that it's still not as up to date as mirror.jenkins-ci.org :)

halkeye:
oh i was talking about the new mirror

PR #211, opened 2020-05-18
Overide default jenkins ingress api version

olblak:


charts/jenkins/values.yaml
```diff
@@ -12,6 +12,8 @@ jenkins:
     testEnabled: false
     image: jenkins/jenkins
     imageTag: 2.236-jdk11
+    ingress:
+      apiVersion: networking.k8s.io/v1beta1
```
timja, line 16: surely this is updated on a newer chart?

olblak: Nope as defined [here](https://github.com/helm/charts/blob/587a6a2ac444dd9b330084783d3aaf77fcfc5779/stable/jenkins/values.yaml#L371) and considering that this chart repository is now deprecated as defined [here](https://github.com/helm/charts/#ddeprecation-timeline) I think it's a temporary solution anyway

PR #217, opened 2020-05-25
Bump Nginx ingress version

olblak:
this podSecurityContext is not (anymore) valid based on helm v3 validation checks


halkeye:
I don't think the context is gone, just moved to a higher tier?

olblak:
> I don't think the context is gone, just moved to a higher tier?

I think it was move in container but this is not something that we can configure with the helm chart

olblak:
I am waiting the end of jenkins UI/hackfest before merging this PR

PR #231, opened 2020-06-03
Add pod labels for network policy

timja:


olblak:
Good catch, I am still wondering if it's the best way to solve this

timja:
> Good catch, I am still wondering if it's the best way to solve this

it's being added automatically in the chart but we're overriding it because we have the multiple nodepools.

olblak:
I used this solution in this [PR](https://github.com/jenkins-infra/charts/pull/232), I couldn't find a better way to handle this as the jenkins helm chart doesn't allow us to define custom podLabels

PR #232, opened 2020-06-04
Cluster/restore

olblak:
This pull request contains all changes that I already executed on the server while redeploying everyhting with helm v3.


config/default/plugin-site.yaml
```diff
@@ -11,12 +11,13 @@ ingress:
     "nginx.ingress.kubernetes.io/cors-allow-origin": "*"
 
   hosts:
+    - host: plugins.origin.jenkins.io
     - host: plugins.jenkins.io
-
+  # www.plugins.io certificate is managed by Fastly
```
timja, line 16:
```suggestion
  # plugins.jenkins.io certificate is managed by Fastly
```

config/default/jenkinsio.yaml
```diff
@@ -46,13 +54,14 @@ ingress:
...
           serviceName: jenkinsio-zh
           servicePort: 80
+  # jenkins.io && www.jenkins.io certificate are managed at Fastly level
   tls:
     - secretName: jenkinsio-tls
       hosts:
-        - jenkins.io
-        - www.jenkins.io
         - jenkins-ci.org
+        - jenkins.io
```
halkeye, line 62: The comment and this line with jenkins.io doesn't match up
was jenkins.io bare supposed to be readded?

helmfile.d/cert-manager.yaml
```diff
@@ -1,11 +1,17 @@
...
+      chart: jetstack/cert-manager
+      namespace: cert-manager
+      wait: true
+      version: v0.15.1
+      set:
+        - name: installCRDs
+          value: true
+
     - name: acme
       chart: ../charts/acme
```
halkeye, line 12: I thought this one inherits from cert-manager. Should it be both of them in the helmfile?

config/default/jenkinsio.yaml
olblak: I updated the comment, only www is now managed by Fastly, jenkins.io was initialy configured on Fastly but we don't need it anymore since we use www.jenkins.io

helmfile.d/cert-manager.yaml
olblak: The idea is to manage crd as part of the helm release, [doc](https://cert-manager.io/docs/installation/kubernetes/#steps)

olblak:
infra.ci.jenkins.io is the last service that stilll needs to be re-deploy before automated deployment can be re-enabled

olblak:
@timja @halkeye Last chance to review (especially two last commits) before I merge and reenable automated deployment

config/default/jenkins-infra.yaml
```diff
@@ -49,19 +54,24 @@ jenkins:
...
+                        jenkinsUrl: "http://jenkins-infra:8080"
                         maxRequestsPerHostStr: "300"
                         name: "kubernetes"
                         namespace: "jenkins-infra"
                         podRetention: "Never"
                         serverUrl: "https://kubernetes.default"
+                        podLabels:
+                          # Required to be jenkins/<helm-release>-jenkins-slave as definede here
+                          # https://github.com/helm/charts/blob/ef0d749132ecfa61b2ea47ccacafeaf5cf1d3d77/stable/jenkins/templates/jenkins-master-networkpolicy.yaml#L27
+                          - key: "jenkins/jenkins-infra-jenkins-slave"
```
timja, line 67: is it worth overriding the componentName so its agent instead of slave?

olblak: I'll do before merging

olblak:
Which part? Maybe I can quickly provide some explaination

halkeye:
Oh I know of it, I just havnt used network policies yet so I don't have much to say

PR #242, opened 2020-06-08
Add repositories for each helmfile as needed so it passes lint in a clean environment

halkeye:
Signed-off-by: Gavin Mogan <git@gavinmogan.com>

### DIFF
--- a/helmfile.d/cert-manager.yaml
+++ b/helmfile.d/cert-manager.yaml
@@ -1,3 +1,6 @@
+repositories:
+  - name: jetstack
+    url: https://charts.jetstack.io
 releases:
     - name: cert-manager
       chart: jetstack/cert-manager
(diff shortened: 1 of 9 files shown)


olblak:
Normally with helm v2 we used this configuration file [helmfile.d/00-repositories.yaml](https://github.com/jenkins-infra/charts/blob/master/helmfile.d/00-repositories.yaml) to define helm repositories but I don't see a good way to do it without duplicating repository configuration in every helmfile at

halkeye:
@olblak the advantage of doing it per helmfile.d, is no longer worry about collisions, and when you delete a helm file, that repo won't be re-added.

helm3 just no longer includes stable by default

I'm not sure why linting worked in the past for things like loki cause the repositories don't seem to be loaded early enough. I wonder if lint or helmfile changed.

PR #251, opened 2020-06-13
Bump plugin site version to 153-290e61 (better sentry errors)

halkeye:


halkeye:
oooh, you two are fast

PR #254, opened 2020-06-16
INFRA-2639 Add plugin-site to infra ci

timja:


halkeye:
Swwweeeeery

config/default/jenkins-infra.yaml
```diff
@@ -119,6 +124,24 @@ jenkins:
...
+                            id('2019081603')
+                            scanCredentialsId('github-access-token')
+                            repoOwner('jenkins-infra')
+                            repository('plugin-site')
+                            includes('master')
+                          }
+                        }
+                        factory {
+                          workflowBranchProjectFactory {
+                            scriptPath('Jenkinsfile')
```
timja, line 141: Thinking about this we might want a different jenkinsfile so it doesn’t get picked up by the other Jenkins instanced that don’t support k8s?

I’m off to bed now @halkeye Feel free to merge / amend if you want to tinker

PR #350, opened 2020-07-22
Disable release.ci config autoreload

olblak:

While INFRA-2674 is not done otherwise each I manually update the job configuration, my manual change is reverted

### DIFF
--- a/config/default/jenkins-release.yaml
+++ b/config/default/jenkins-release.yaml
@@ -15,6 +15,9 @@ jenkins:
   master:
     image: jenkins/jenkins@sha256
     imageTag: dfe766e539dc2f47600229c7ed5bb7950b34d2c61f1ae240ef9393cbc0df6e68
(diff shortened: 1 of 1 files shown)


config/default/jenkins-release.yaml
```diff
@@ -15,6 +15,9 @@ jenkins:
   master:
     image: jenkins/jenkins@sha256
     imageTag: dfe766e539dc2f47600229c7ed5bb7950b34d2c61f1ae240ef9393cbc0df6e68
+    sidecars:
+      configAutoReload:
```
timja, line 19: it'll still reload on any changes that cause a reboot

olblak: I know :p

olblak: I am working on it, so it's fine as long as we don't reboot the service

PR #356, opened 2020-07-23
Use statefulset for prometheus alertmanager

olblak:

Recent prometheus update was blocked because a volume was already attached the exist container, switch alertmanager to a statefulset better handle such usecases


olblak:
I am merging this right now as I already manually did the change to unblock infra.ci

PR #393, opened 2020-08-08
Add more projects to infra ci

halkeye:
the dockerhub credentials need to be copied over

https://github.com/jenkins-infra/pipeline-library/blob/c88ea135745ba7f3cd45f2a5a3590d7dccfb1606/vars/infra.groovy#L14

I *think* the default jenkins file will still  work since it should auto find a node since there are none with linux && high mem, but we will see

Might also need to setup a dind setup like i did or something
https://github.com/halkeye-docker/docker-jenkins/blob/master/jenkins.yaml#L52

### DIFF
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -148,6 +148,24 @@ jenkins:
                           }
                         }
                       }
+                  - script: >
+                      multibranchPipelineJob('jenkins-wiki-exporter') {
+                        displayName "Wiki Exporter"
+                        branchSources {
+                          github {
+                            id('2019081603')
+                            scanCredentialsId('github-access-token')
+                            repoOwner('jenkins-infra')
+                            repository('jenkins-wiki-exporter')
+                            includes('master')
+                          }
+                        }
+                        factory {
+                          workflowBranchProjectFactory {
+                            scriptPath('Jenkinsfile')
+                          }
+                        }
+                      }
               ldap-settings: |
                 jenkins:
                   securityRealm:
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -166,6 +166,24 @@ jenkins:
                           }
                         }
                       }
+                  - script: >
+                      multibranchPipelineJob('custom-distribution-service') {
+                        displayName "Custom Distribution Service"
+                        branchSources {
+                          github {
+                            id('2019081603')
+                            scanCredentialsId('github-access-token')
+                            repoOwner('jenkinsci')
+                            repository('custom-distribution-service')
+                            includes('master')
+                          }
+                        }
+                        factory {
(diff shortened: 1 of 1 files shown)


config/default/jenkins-infra.yaml
```diff
@@ -143,6 +143,24 @@ jenkins:
...
+                            id('2019081603')
+                            scanCredentialsId('github-access-token')
+                            repoOwner('jenkins-infra')
+                            repository('jenkins-wiki-exporter')
+                            includes('master')
+                          }
+                        }
+                        factory {
+                          workflowBranchProjectFactory {
+                            scriptPath('Jenkinsfile')
```
timja, line 160: Might want this to be Jenkinsfile_k8s so that ci.jenkins.io won’t pick it up? That’s if you add any k8s specific stuff to it

halkeye: > I think the default jenkins file will still work since it should auto find a node since there are none with linux && high mem, but we will see

> Might also need to setup a dind setup like i did or something
https://github.com/halkeye-docker/docker-jenkins/blob/master/jenkins.yaml#L52

I'm hoping we don't need to make a separate Jenkins file.

Since creating this PR though, I'm wondering if we can use img to build images without dind

https://github.com/genuinetools/img#running-with-docker

timja: I’ve never heard of img, kaniko is usually the popular one although ive never used it

halkeye: Kaniko is a root-less docker replacement, so including running images
img is docker apis, allows to build, but not run without needing to change dockerfiles and stuff. In theory at least. I've been itching to try it, so hopefully will get to this weekend

olblak: I'm hoping we don't need to make a separate Jenkins file.
While I like the idea very much, I don't see how we can do that as the Jenkins agent must be differently configured.

Regarding img versus kaniko they both have pros and cons and are both quite popular so I am fine to pick one.

Whatever we select we need a secret in Jenkins containing `.docker/config.json`, any better idea than `config.json < echo "$DOCKER_CONFIG" | base64 -d`

timja: @olblak you can use some macros to convert the json file to base64 in jcasc:
https://github.com/jenkinsci/configuration-as-code-plugin/tree/master/demos/credentials#credentials-plugin

olblak: > @olblak you can use some macros to convert the json file to base64 in jcasc
Nice it will definitely help to review secret changes

halkeye:
Adding wiki exporter and custom distribution service
https://github.com/jenkinsci/custom-distribution-service/pull/131
https://github.com/jenkins-infra/jenkins-wiki-exporter/pull/129

They both need https://github.com/jenkins-infra/pipeline-library/pull/156

PR #427, opened 2020-08-18
Move updatecli to jenkins-infra/updatecli

olblak:
This pull request move updatecli configuration files from this repository to jenkins-infra/updatecli, this will allow us to use it for more git repository than just jenkins-infra/charts.


halkeye:
I like the seperation of the job. But I like the configs being with the charts so they get updated together.

The big one for me is I'd love to see this run As seperate user so it's easier to sort in email

olblak:
> I like the separation of the job. But I like the configs being with the charts so they get updated together.
While I also like the config being closed to helm chart, I am envisioning to run 'updatecli' on different timeframe like weekly and hourly for now, also I would like to use for more than jenkins-infra/charts.
What do you think?

I am fine to change the approach to add updatecli file to multipe repositories.

> The big one for me is I'd love to see this run As seperate user so it's easier to sort in email
Yep, we already talked about that and this is definitely something that I have to improve

PR #449, opened 2020-08-26
Add autoupdate of geoip data

timja:
Uses https://hub.docker.com/r/maxmindinc/geoipupdate

I've tested this in the mirrorbits2 namespace in the cluster (now deleted), data is updated

### DIFF
--- a/charts/mirrorbits/templates/deployment.yaml
+++ b/charts/mirrorbits/templates/deployment.yaml
@@ -25,18 +25,6 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
-        - name: crond
-          image: "{{ .Values.image.mirrorbits.repository }}:{{ .Values.image.mirrorbits.tag }}-cron"
-          imagePullPolicy: {{ .Values.image.mirrorbits.pullPolicy }}
-          securityContext:
-            allowPrivilegeEscalation: false
-          resources:
-            {{- toYaml .Values.resources.mirrorbits | nindent 12 }}
-          volumeMounts:
-            {{ if $.Values.repository.persistentVolumeClaim.enabled -}}
-            - name: binary
-              mountPath: /srv/repo
-            {{- end }}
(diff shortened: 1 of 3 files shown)


charts/mirrorbits/templates/deployment.yaml
```diff
@@ -25,18 +25,6 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
-        - name: crond
```
timja, line 28: I moved the order as it's annoying that the default container is crond when you don't specify anything 👅 

like 10 extra characters to exec into the pod.

charts/mirrorbits/values.yaml
```diff
@@ -85,6 +85,15 @@ tolerations: []
 
 affinity: {}
 
+geoipupdate:
+  image:
+    repository: maxmindinc/geoipupdate
```
halkeye, line 90: oh nice, there's an official one?
I've used https://github.com/halkeye-helm-charts/tautulli/blob/master/values.yaml#L40-L48 before

charts/mirrorbits/values.yaml
```diff
@@ -85,6 +85,15 @@ tolerations: []
...
+    tag: v4.3
+  #account_id:
```
halkeye, line 92: I think this needs to be set to empty for linting purposes.

charts/mirrorbits/values.yaml
timja: Yeah since April: https://github.com/maxmind/geoipupdate/issues/24

charts/mirrorbits/values.yaml
timja: `1 chart(s) linted, 0 chart(s) failed`

the chart doesn't work with no values provided anyway...

needs azure keys and ingress to be configured iirc

halkeye: fair, I wasn't certain if you could b64enc nil. I know it complains when dicts are undefined

PR #487, opened 2020-09-11
Fix geoipdata race condition

timja:
Geoip data must exist before mirrorbits starts up

### DIFF
--- a/charts/mirrorbits/templates/deployment.yaml
+++ b/charts/mirrorbits/templates/deployment.yaml
@@ -83,6 +83,25 @@ spec:
           envFrom:
           - secretRef:
               name: {{ include "mirrorbits.fullname" . }}-geoipupdate
+      initContainers:
+        # we need to ensure the geoip data is in place before mirrorbits starts
+        - name: geoipupdate-init
+          image: "{{ .Values.geoipupdate.image.repository }}:{{ .Values.geoipupdate.image.tag }}"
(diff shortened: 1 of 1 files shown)


charts/mirrorbits/templates/deployment.yaml
```diff
@@ -83,6 +83,25 @@ spec:
...
+          securityContext:
+            allowPrivilegeEscalation: false
+          volumeMounts:
+            - name: geoipdata
+              mountPath: /usr/share/GeoIP
+          resources:
+            {{- toYaml .Values.resources.mirrorbits | nindent 12 }}
+          env:
+            - name: GEOIPUPDATE_EDITION_IDS
+              value: {{ .Values.geoipupdate.editions }}
```
halkeye, line 100: I would set GEOIPUPDATE_FREQUENCY  explicitly just in case it ever gets set in secret ref for some dumb reason

```suggestion
              value: {{ .Values.geoipupdate.editions }}
            - name: GEOIPUPDATE_FREQUENCY
              value: "0"
```

halkeye (approved):

charts/mirrorbits/templates/deployment.yaml
timja: hmm I think it would still win if set in the secret, the issue would be really obvious as the container wouldn't exit.

I'd prefer not to add.

PR #641, opened 2020-11-23
Remove deprecated application from publick8s

olblak:
Remove some deprecated helm charts from publick8s which were used for Jenkins election 2019.
Once removed here, I'll manually delete them from the cluster.


olblak:
The oauth-proxy helm chart is not supported anymore -> https://github.com/helm/charts/blob/master/stable/oauth2-proxy/
The dex helm chart is not supported anymore -> https://github.com/helm/charts/tree/master/stable/dex
In the current state, the poll helm chart won't be used anymore

garethjevans (approved):

timja (approved):

PR #644, opened 2020-11-24
Do not check datadog docker image tag

olblak:

The default behavior for Datadog helm chart is to check the docker image tag used which IMHO is a very good idea but I need a little bit more time to build a custom docker image with an appropriate tag 

```
==> Linting /tmp/253114755/datadog/2.5.4/datadog/datadog/datadog

[ERROR] templates/: template: datadog/templates/_helpers.tpl:16:12: executing "check-version" at <semverCompare "^6.19.0-0 || ^7.19.0-0" $version>: error calling semverCompare: Invalid Semantic Version

```


olblak:
I am going to merge this PR as I would like to start monitoring if we can download distribution packages before the next weekly release

PR #691, opened 2020-12-11
UpdateCLI runs every 30 min, regular apply/lint happens on all branches

halkeye:
Unless your watching the CI jobs, its pretty hard to see when something was changed (for example chatbot).

Split the updatecli job and the helmfile job so one is cron based, one is commit based.

Could setup helmfile to pin chart versions and get updatecli to bump the chart version.

### DIFF
--- a/Jenkinsfile_updatecli_k8s
+++ b/Jenkinsfile_updatecli_k8s
@@ -19,6 +19,7 @@ pipeline {
   }
 
   triggers {
+    cron 'H/30 * * * *'
   }
 
   stages {
@@ -39,35 +40,25 @@ pipeline {
         }
       }
     }
-    stage('Prepare Environment'){
+    stage('Check Configuration Update') {
       when { branch 'master' }
+      environment {
+        UPDATECLI_GITHUB_TOKEN  = credentials('updatecli-github-token')
+      }
       steps {
-        container('helmfile'){
-          sh 'mkdir -p $HOME $HOME/.config'
+        container('updatecli') {
+          sh 'updatecli diff --config ./updateCli/updateCli.d --values ./updateCli/values.yaml'
         }
       }
     }
-    stage('Test Lint'){
+    stage('Apply Configuration Update') {
       when { branch 'master' }
-      steps {
-        container('helmfile'){
-          sh 'helmfile -f clusters/publick8s.yaml lint'
-        }
+      environment {
+        UPDATECLI_GITHUB_TOKEN  = credentials('updatecli-github-token')
       }
-    }
-    stage('Diff'){
-      when { branch 'master' }
       steps {
-        container('helmfile'){
-          sh 'helmfile -f clusters/publick8s.yaml diff --suppress-secrets'
-        }
-      }
-    }
-    stage('Apply'){
-      when { branch 'master' }
-      steps {
-        container('helmfile'){
-          sh 'helmfile -f clusters/publick8s.yaml apply --suppress-secrets'
+        container('updatecli') {
+          sh 'updatecli apply --config ./updateCli/updateCli.d --values ./updateCli/values.yaml'
         }
       }
     }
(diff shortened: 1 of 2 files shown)


Jenkinsfile_updatecli_k8s
```diff
@@ -0,0 +1,66 @@
...
+
+  stages {
+    stage('Yaml Lint') {
+      steps {
+        container('yamllint') {
+          sh 'yamllint --config-file yamllint.config config/'
+        }
+      }
+    }
+    stage('Init Secrets') {
```
olblak, line 33: Not Needed for updatecli

Jenkinsfile_updatecli_k8s
```diff
@@ -0,0 +1,66 @@
...
+    timeout(time: 30, unit: 'MINUTES')
+    disableConcurrentBuilds()
+  }
+
+  triggers {
+    cron 'H/30 * * * *'
+  }
+
+  stages {
+    stage('Yaml Lint') {
```
olblak, line 26: Not Needed for updatecli

Jenkinsfile_updatecli_k8s
```diff
@@ -0,0 +1,66 @@
+pipeline {
+  agent {
+    kubernetes {
+      label 'helmfile'
+      yamlFile 'PodTemplates.yaml'
+      inheritFrom 'jnlp-linux'
+    }
+  }
+  environment {
```
olblak, line 9: Not Needed for updatecli

halkeye:
So not a dumb idea? I'll totally clean this up tomorrow

olblak:
> Unless your watching the CI jobs, its pretty hard to see when something was changed (for example chatbot).
I definitely agree with you even though I didn't find a good approach to solve this with helmfile

IMHO the Helmfile apply should be run on a regular basis to ensure that any manual configuration change is reverted.
I agree that it's hard to catch when something changed but I don't think running Helmfile from a different job will help.
I don't think a Jenkins job could detect if Helmfile apply changes or not because the command doesn't return a different RC in c are of a change.

I have been wondering if it's not something that change be detected by prometheus by watching helm release changes.

halkeye:
>  IMHO the Helmfile apply should be run on a regular basis to ensure that any manual configuration change is reverted.

Maybe less often? Like once a day?

I'm going to check Loki tomorrow, was just trying to figure out why chatbot wasn't deploying

olblak:
Ok then let split the two and run updatecli every 15min and helmfile once a day

Ideally, I would like to get something similar to puppet notifications` in IRC like 

```
00:33 < infra-butler> Puppet production run for bounce changed at Fri Dec  4 00:32:59 2020.
```

halkeye:
So IRC notifications for helmfile? Is that available on infra.ci?

I'll get the changes cleaned up asap

olblak:
> So IRC notifications for helmfile? Is that available on infra.ci?
> 
> I'll get the changes cleaned up asap

The closest solution that I could think of would be to use helmfile hooks base on prepare event 
https://github.com/roboll/helmfile#hooks

olblak:
Now updatecli is only executed when the job is triggered with a cronjob as defined [here](https://github.com/jenkins-infra/charts/blob/3c316c7e34b3f1079233533995c8c1e8c5379d4a/Jenkinsfile_k8s#L64) by @dduportal 
 
So I am closing this PR

dduportal:
MMmh the "conditional execution when triggered by a cron" had been a first step towards error-less helmfile apply.

But it feels like that this PR, which object is to split the updatecli and helm in 2 separated pipelines, should be the next step.


What do you think about having a pipeline library for this? E.g. defining an updateCli process (input parameter: frequency, and a collection of branches) for this repository. it would simplify a lot of our pipeline and would provide a kind of "dependanbot" for us.

PR #791, opened 2021-01-25
Add X-Frame-Options DENY to stop the pestering

jvz:
https://infosec.mozilla.org/guidelines/web_security#x-frame-options

Although the site doesn't use iframes or have accounts, various security scanners and their associated operators continue to complain about this.

#### Is this a new chart
- [ ] `CODEOWNERS` is updated
- [ ] If secrets are needed, `.sops.yaml` must be updated with appropriated public gpg key ID, and full public gpg key must be provided to @olblak

#### What this PR does / why we need it:

Adds the X-Frame-Options: DENY header by default to nginx pages. Various script kiddies continue to spam the Jenkins Security Team about this, and it's harmless to enable.

#### Which issue this PR fixes

  - fixes # [INFRA-XXXX](https://issues.jenkins-ci.org/projects/INFRA/issues/INFRA-XXXX)

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. `[stable/mychartname]`)

@daniel-beck

### DIFF
--- a/charts/jenkinsio/templates/nginx-configmap.yaml
+++ b/charts/jenkinsio/templates/nginx-configmap.yaml
@@ -30,6 +30,8 @@ data:
             rewrite ^/$ https://www.jenkins.io/$lang$1;
           }   
       }
+      
+      add_header X-Frame-Options "DENY";
 
       location ~* \.(?:css|js|woff|eot|svg|ttf|otf|png|gif|jpe?g) {
         expires 2d;


markyjackson-taulia (approved):

MarkEWaite (approved):
Not an nginx expert, but no objection to adding that header.

olblak (approved):

timja (approved):
go away script kiddies

charts/jenkinsio/templates/nginx-configmap.yaml
```diff
@@ -30,6 +30,8 @@ data:
             rewrite ^/$ https://www.jenkins.io/$lang$1;
           }   
       }
+      
+      add_header X-Frame-Options "DENY";
```
timja, line 34: do you need to enable it on the chinese version too?

https://github.com/jvz/charts/blob/patch-1/charts/jenkinsio/templates/zh-configmap.yaml

olblak (approved):

charts/jenkinsio/templates/nginx-configmap.yaml
jvz: Perhaps once Chinese script kiddies start to pester us about that subdomain, too :P

jvz:
I don't see the header showing up yet; was that the right config?

halkeye:
I see it, but its only showing up once you get redirected to https://www.jenkins.io

curl -v https://jenkins.io/ -L 2>&1 | grep -i frame
< x-frame-options: DENY
  .video_105 iframe {
<iframe allow='accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture' frameborder='0' height='621' src='https://www.youtube.com/embed/_MXtbjwsz3A' width='1261'></iframe>

jvz:
Oh, good enough I hope! Thanks for confirming

PR #821, opened 2021-02-02
✨ Execute the updateCli's pipeline steps only when the pipeline is triggered by the daily cron routine

dduportal:
As per https://www.jenkins.io/doc/book/pipeline/syntax/#when, the goal is to NOT fail the pipeline when we need to apply configuration from helmfile/values to the cluster.

The idea of this PR is to make sure that updateCli only runs on the daily routine.

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -40,7 +40,13 @@ pipeline {
       }
     }
     stage('Check Configuration Update') {
-      when { branch 'master' }
+      when {
+        allOf {
+          branch 'master'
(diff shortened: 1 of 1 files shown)


markyjackson-taulia (approved):

dduportal:
The pipeline has the 2 following lines:

```shell
# ...
Stage "Check Configuration Update" skipped due to when conditional
# ...
Stage "Apply Configuration Update" skipped due to when conditional
# ...
```

which is the expected result \o/

olblak (approved):
\o/

PR #859, opened 2021-02-16
chore: switch infra.ci to use jenkinsciinfra/jenkins-weekly image

garethjevans:
This PR switches infra.ci to use the prebuilt docker image for jenkins (including all plugins) that is defined at: https://github.com/jenkins-infra/docker-jenkins-weekly

### DIFF
--- a/charts/jenkins/values.yaml
+++ b/charts/jenkins/values.yaml
@@ -24,8 +24,8 @@ jenkins:
       readinessProbe:
         initialDelaySeconds: 120
     testEnabled: false
-    image: jenkins/jenkins
-    tag: 2.286-jdk11
+    image: jenkinsciinfra/jenkins-weekly
+    tag: 0.0.7-2.286
     ingress:
       apiVersion: networking.k8s.io/v1beta1
     runAsUser: 1000
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -465,11 +465,17 @@ jenkins:
               globalMatrix:
                 permissions:
                   - "Overall/Administer:admins"
+<<<<<<< HEAD
                   - "Overall/SystemRead:authenticated"
                   - "Overall/Read:authenticated"
                   - "Job/Read:authenticated"
                   - "Job/Build:authenticated"
                   - "Job/ExtendedRead:authenticated"
+                  - "Overall/SystemRead:all"
+                  - "Overall/Read:all"
+                  - "Job/Read:all"
+                  - "Job/Build:all"
+                  - "Job/ExtendedRead:all"
         system-settings: |
           jenkins:
             disabledAdministrativeMonitors:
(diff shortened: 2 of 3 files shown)


markyjackson-taulia (dismissed):

charts/jenkins/values.yaml
```diff
@@ -24,8 +24,8 @@ jenkins:
       readinessProbe:
         initialDelaySeconds: 120
     testEnabled: false
-    image: jenkins/jenkins
-    tag: 2.279-jdk11
+    image: jenkinsciinfra/jenkins-weekly@sha256
+    tag: 81e0b4dd49ecd529104c2b2244bb3cf0d927b07d2ae99a53b4ff56b77eea6a1
```
timja, line 28: why not use a human readable tag?

garethjevans: at the moment, that docker image is not released with a human readable tag - its the same with `jenkinsciinfra/jenkins-lts`

dduportal: Specifying a digest for a Docker image is also an additional layer of safety, which is nice to have for a sensitive instance as infra.ci

markyjackson-taulia (approved):

charts/jenkins/values.yaml
olblak: > Specifying a digest for a Docker image is also an additional layer of safety, which is nice to have for a sensitive instance as infra.ci

we used digest for a while, I find them quite difficult to review which why I recently switch back to specific tag

olblak: I think it would have been useful to bring the context, there is this repository https://github.com/jenkins-infra/docker-jenkins-weekly that builds a custom Jenkins docker image with the plugins that we need. It doesn't use releases yet https://github.com/jenkins-infra/docker-jenkins-weekly so the tags available are "edge" or "master". While I understand that it's better in terms of reproducibility to use digest, I find it more convenient to review image tag changes, but that's a personal opinion.

dduportal: Fair, with some context it explains clearly the intent to have human readable tags (e.g. to ease review).

This might be a good use case for using the "default docker build" process in combination with @garethjevans 's `inspect` tool in the future (but no priority and out of scope of this PR of course).

garethjevans: I was wondering if we could supply a kind of `tag template` so that we can control the tag that is given to the docker images with the default pipeline library build?

garethjevans: e.g. could we use the `jenkins version` somewhere in the name used for the tag

garethjevans: switched to a human readable tag

olblak (approved):

dduportal (approved):
LGTM 👍

PR #911, opened 2021-03-02
Tune JVM for infra.ci: resources and JAVA_OPTS

dduportal:
The goal of this PR is to improve the resources usage and performance of infra.ci:

- After #908 , we received a first report from Health Advisor that points to <https://support.cloudbees.com/hc/en-us/articles/204859670-Java-Heap-settings-best-practice>. This PR adds the proposed java opts flags adapted for the JDK 11

Capture d’écran 2021-03-02 à 09 49 52

- Set the initial resources requested for Jenkins to the same as the limit to ensure that the JVM has enough resources at startup to ensure a fast and efficient startup. The goal is to start from there and then improve monitoring of the JVM usage to ensure that we can add jobs efficiently, restart the instance fast, and diagnose the "LDAP" timeout issues.

- After #908 where an administrative monitor is dismissed as code, this PRs adds a new system config to dismiss another monitor which was back to the menu

- Minor comment from https://github.com/jenkins-infra/charts/pull/907#discussion_r584851396

### DIFF
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -448,6 +448,10 @@ jenkins:
                   - "Job/Read:all"
                   - "Job/Build:all"
                   - "Job/ExtendedRead:all"
+        system-settings: |
+          jenkins:
+            disabledAdministrativeMonitors:
+              - "jenkins.security.QueueItemAuthenticatorMonitor"
     ingress:
       enabled: true
       hostName: infra.ci.jenkins.io
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -9,7 +9,14 @@ jenkins:
       namespaceLabels:
         name: "jenkins-infra"
   controller:
-    javaOpts: "-XshowSettings:vm -XX:+AlwaysPreTouch -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled -XX:+DisableExplicitGC"
+    resources:
+      limits:
+        cpu: "2"
+        memory: "4Gi"
+      requests:
+        cpu: "2"
+        memory: "4Gi"
+    javaOpts: "-XshowSettings:vm -XX:+AlwaysPreTouch -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled -XX:+DisableExplicitGC -XX:MaxRAM=4g"
     JCasC:
       enabled: true
       defaultConfig: false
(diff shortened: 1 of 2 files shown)


dduportal:
Putting the PR on-hold, waiting for the weekly release

olblak (dismissed):
In the future, we could try to fine tune `-XX:MaxRAMPercentage` but I think for now it, this PR is ready to go

oleg-nenashev (commented):
I doubt that 4Gb is enough for production use. Infra CI runs quite heavy payloads. Do we have some resource usage stats?

config/default/jenkins-infra.yaml
```diff
@@ -9,6 +9,14 @@ jenkins:
...
         name: "jenkins-infra"
   controller:
+    resources:
+      limits:
+        cpu: "2"
+        memory: "4Gi"
+      requests:
+        cpu: "2"
+        memory: "4Gi"
+    javaOpts: "-XshowSettings:vm -XX:+AlwaysPreTouch -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled -XX:+DisableExplicitGC -XX:MaxRAM=4g"
```
olblak, line 19:
```suggestion
    # Following javaOpts assumes that we are using the jdk-11
    javaOpts: "-XshowSettings:vm -XX:+AlwaysPreTouch -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled -XX:+DisableExplicitGC -XX:MaxRAM=4g"
```

dduportal:
@oleg-nenashev 

> I doubt that 4Gb is enough for production use. Infra CI runs quite heavy payloads. Do we have some resource usage stats?

The current applied values for infra.ci come from the default's of the official helm chart: https://github.com/jenkinsci/helm-charts/blob/main/charts/jenkins/values.yaml#L61

When you reference heavy workload, which one are thinking in particular? (Newbie question but the only things that I know of the infra.ci workload and not really heavy: helmfile tasks, or building docker images for the few jobs I know)

We currently have the following metrics: <https://grafana.publick8s.jenkins.io/d/kubernetes_app_metrics/kubernetes-app-metrics?orgId=1&refresh=30s&var-namespace=jenkins-infra&var-container=&from=now-2d&to=now> that might or might not be enough.

Keep in mind that the scope of this PR is to "get started": given the actual values, it tries to make the best of it (e.g. giving all the available resources, ensure there is no OOM and improve JVM usage). The goal would be to iterate over this settings by checking metrics for the next steps, WDYT?

olblak (approved):

markyjackson-taulia (approved):

PR #951, opened 2021-03-11
feat: add chart for captain-hook as a webhook relay for infra

garethjevans:


timja (approved):
this should probably live in jenkinsci and have it's helm chart in https://github.com/jenkinsci/helm-charts but good for now I guess

oleg-nenashev (commented):
Have you considered using https://github.com/stakater/GitWebhookProxy ? Not against using a new implementation, but there might be some maintenance overhead

timja:
> Have you considered using [stakater/GitWebhookProxy](https://github.com/stakater/GitWebhookProxy) ? Not against using a new implementation, but there might be some maintenance overhead

I guess the question is does what does that proxy do if jenkins is down? part of captain-hook is storing them and only forwarding them if Jenkins is up

garethjevans:
@oleg-nenashev I wasn't aware of that project.  But from a brief look it does appear to be slightly different.  It doesn't make any attempt to redeliver the hook if jenkins is unavailable for a period of time

I've also added a CRD to aid debuging e.g.

```console
kubectl get hooks
```

will display the webhook events its currently processing.

olblak:
> this should probably live in jenkinsci and have it's helm chart in https://github.com/jenkinsci/helm-charts but good for now I guess

I would definitely see it under the jenkinsci organisation, it will bring visibility on this tool and maybe attract contributors and why not also integrate it in the jenkinsci helm chart once stable enough

garethjevans:
@timja totally agree with the location of the chart etc, just want to do a little more testing against some real repos to see how it handles the restarts.

Need to speak to @olblak about using a different TLS host and firewall rules to allow requests in.

olblak (approved):

markyjackson-taulia (approved):

oleg-nenashev (approved):

PR #968, opened 2021-03-16
[NFRA-2903] Hotfix: fix infra.ci LDAP settings

dduportal:
This PR fixes the infra.ci issues when using LDAP authentication.

- Define a search base explictely to ensure that the LDAP requests emitted when authenticating are not HUGE. This settings comes from ci.jenkins.io's JCasc export.
- Switches the RBAC setting from the group `all` to the group `authenticated` to avoid HUGE requests on the LDAP
- Nice to have: fine tune the headdump in JAVA_OPTS

### DIFF
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -16,7 +16,9 @@ jenkins:
       requests:
         cpu: "2"
         memory: "4Gi"
-    javaOpts: "-XshowSettings:vm -XX:+AlwaysPreTouch -XX:+UseStringDeduplication -XX:+ParallelRefProcEnabled -XX:+DisableExplicitGC -XX:MaxRAM=4g"
+    javaOpts: >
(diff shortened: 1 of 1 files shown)


markyjackson-taulia (approved):

garethjevans (approved):

oleg-nenashev (approved):
looks reasonable

config/default/jenkins-infra.yaml
```diff
@@ -453,11 +454,11 @@ jenkins:
               globalMatrix:
                 permissions:
                   - "Overall/Administer:admins"
-                  - "Overall/SystemRead:all"
-                  - "Overall/Read:all"
-                  - "Job/Read:all"
-                  - "Job/Build:all"
-                  - "Job/ExtendedRead:all"
+                  - "Overall/SystemRead:authenticated"
```
olblak, line 457: what is the authenticated? Is it supposed to be a ldap group?

dduportal: No, it's a virtual group in Jenkins, which maps to any authenticated user through any of the defined realm. It's been added following a tip from Daniel in IRC, which make sense in this instance's case.

PR #974, opened 2021-03-17
Chore/switch ci sidecars to polling

dduportal:
This PR introduces the following changes:

* Chore: set infra.ci and release.ci Jenkins' sidecars to polling mode instead of watching.
  - The goal is to avoid missing configmap update events.
  - Ref. https://github.com/hmcts/cnp-flux-config/pull/8674 and https://github.com/kiwigrid/k8s-sidecar/issues/85
  - Drawback: deletion of configmap is not detected with this change. It means that removing top-level entries on the JCasc value config will require a pod deletion.
* Bugfix: with #972 , I introduced an issue with the indentation of the advisor plugin configuration.
* Chore: increase the datadog chart timeout as the builds are often failing these days, while applying the datadog step

### DIFF
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -463,6 +463,17 @@ jenkins:
           jenkins:
             disabledAdministrativeMonitors:
               - "jenkins.security.QueueItemAuthenticatorMonitor"
+    sidecars:
+      configAutoReload:
+        env:
+          # https://github.com/kiwigrid/k8s-sidecar#configuration-environment-variables
+          - name: METHOD
+            # Polling mode (instead of watching kube API)
+            value: "SLEEP"
+          # https://github.com/kiwigrid/k8s-sidecar#configuration-environment-variables
+          - name: SLEEP_TIME
+            # Time in seconds between two polls
+            value: "60"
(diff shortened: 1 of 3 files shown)


timja (dismissed):

dduportal:
@timja the duration of 300 looked a lot. I changed to 60 to ensure the configmap are checked often (and to avoid waiting too much when deploying). WDYT?

markyjackson-taulia (dismissed):

olblak:
> @timja the duration of 300 looked a lot. I changed to 60 to ensure the configmap are checked often (and to avoid waiting too much when deploying). WDYT?

Is this seconds or minutes?

dduportal:
> Is this seconds or minutes?

As per the documentation at https://github.com/kiwigrid/k8s-sidecar, it is in seconds:

> SLEEP_TIME
    description: How many seconds to wait before updating config-maps/secrets when using SLEEP method.
    required: false
    default: 60
    type: integer

Let me add a comment on the values to help on this

olblak:
IMHO 60 is fine, we can still adjust that value later. Thanks for this PR

olblak (approved):

PR #982, opened 2021-03-22
Update trusted.ci VM IP to allow incoming LDAP connection

dduportal:


garethjevans (approved):

halkeye:
ooooh, backwards. I was trying to figure out why trusted.ci needed to allow incoming from ldap.

To be clear, this is allowing trusted to talk to ldap right?

dduportal:
@halkeye that is correct. The trusted VM being an EC2 instance with no further tuning that I'm aware of, its public IP changes when restarted (which happened today due to the VM being non reachable). There is room for improvement to avoid this scenario though (an Elastic IP, or automation for this VM, etc.)

PR #1017, opened 2021-04-01
chore: switch release.ci to use prebuilt image

garethjevans:


charts/jenkins/values.yaml
```diff
@@ -3,7 +3,7 @@ clusterAdminEnabled: false
 jenkins:
   rbac:
     create: true
-    readSecrets: false
+    readSecrets: true
```
timja, line 6: why?

timja (approved):

charts/jenkins/values.yaml
garethjevans: to allow the kubernetes-credentials-provider plugin to read secrets and turn them into jenkins credentials

timja (approved):

charts/jenkins/values.yaml
olblak: What's the reason to no use the credentials defined from the configuration as code plugin? 
Several months ago we stopped using kubernetes secrets in favor of jcasc

garethjevans: This gives us the option of using kubernetes external secrets with the credentials provider to load them directly from a cloud hosted secrets manager (rather than using the sops option)

garethjevans: I've not moved any secrets or changed how it currently works, just enabled the plugin so we have the option.  Using this approach secret rotation becomes far easier.

olblak (approved):

PR #1030, opened 2021-04-06
chore: switch release.ci over to use websocket config

garethjevans:
brings the config inline with changes made on the UI


olblak (approved):
Thanks @garethjevans for this. I'll merge it once the release is done

PR #1070, opened 2021-04-15
jenkinsio helm chart used a self-contain docker image for the default website

olblak:
The purpose of this pull request is to use a self-contained docker image for the English Jenkins.io website.
In the future the current helm chart use a docker image build from infra.ci using this [Jenkinsfile](https://github.com/jenkins-infra/jenkins.io/blob/master/Jenkinsfile_k8s)
The current workflow relies on updatecli continuously bump the docker image digest as defined [here](https://github.com/jenkins-infra/jenkins.io/blob/master/updatecli/updatecli.d/charts.yaml)


olblak:
If something goes wrong we can still rollback this pull request and rely on the azure file storage approach as before

garethjevans (approved):

dduportal (approved):
LGTM 👍 
LESS CODE LESS VOLUME SHIP IT 🚀

PR #1074, opened 2021-04-15
3x increase helmfile container memory request

MarkEWaite:
Reduce out of memory failures during www.jenkins.io deploy as seen in checks for master branch recently.


halkeye (approved):
YOLO

PR #1105, opened 2021-04-27
Double mirrorbits container usage limit

olblak:
While reviewing mirrorbits resource usage on publickk8s, I saw that the third containers were at their max capacity, because of the IO constraint on the azure file storage below, I would like to first increase the CPU/memory instead of adding more mirrorbits containers.

### DIFF
--- a/config/default/mirrorbits.yaml
+++ b/config/default/mirrorbits.yaml
@@ -5,11 +5,11 @@ replicaCount:
 resources:
   mirrorbits:
     limits:
-      cpu: 1000m
-      memory: 1024Mi
+      cpu: 2000m
+      memory: 2048Mi
     requests:
-      cpu: 1000m
-      memory: 1024Mi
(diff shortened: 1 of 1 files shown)


dduportal (approved):

timja (approved):

dduportal:
Non blocking question: is the memory increase needed (e.g. did you see OOMkill or memory pressure on the node)?

Asking as the CPU make sense given the I/O constraint, but not sure of the memory impact

olblak:
> Non blocking question: is the memory increase needed (e.g. did you see OOMkill or memory pressure on the node)?
> 
> Asking as the CPU make sense given the I/O constraint, but not sure of the memory impact

I haven't seem OOM issues but I would like to monitor how the container when it has more capacity. 
I would also be interested to see if more containers make the service more stable but I prefer to do one change at a time

PR #1124, opened 2021-04-28
(breaking) Finish infra.ci's DSL cleanup

dduportal:
This PR is a follow up of #1123 (read it for context) and introduces the following breaking change on infra.ci's configuration:

- Factorize all multibranch jobs on a single loop-based configuration
- Factorize all gh-org scanning jobs on a single loop-based configuration (keeping 2 separated even if the same organization)

Please note that the "K8S management" job is gonna be re-created under a new URL, which is the breaking change. The older one has been renamed and disabled, and should be deleted only after 1 week (to allow rollback if required)

### DIFF
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -297,67 +297,6 @@ jenkins:
         jobs-settings: |
           jobs:
             - script: >
-                multibranchPipelineJob('k8smgmt') {
-                  displayName "K8s Cluster Management"
-                  description "Current K8s Cluster Management"
-                  branchSources {
-                    github {
-                      id('2019081602')
-                      scanCredentialsId('github-app-infra')
-                      repoOwner('jenkins-infra')
-                      repository('charts')
-                    }
-                  }
-                  factory {
-                    workflowBranchProjectFactory {
-                      scriptPath('Jenkinsfile_k8s')
-                    }
-                  }
-                  orphanedItemStrategy {
-                    discardOldItems {
-                      daysToKeep(1) // Keep removed SCM heads/branch/PRs only for 1 day
-                    }
-                  }
-                  configure { node ->
-                    def traits = node / 'sources' / 'data' / 'jenkins.branch.BranchSource' / 'source' / 'traits'
-                    traits << 'org.jenkinsci.plugins.github__branch__source.BranchDiscoveryTrait' {
-                      strategyId(1)
-                    }
-                    traits << 'org.jenkinsci.plugins.github__branch__source.OriginPullRequestDiscoveryTrait' {
-                      strategyId(1)
-                    }
-                    traits << 'org.jenkinsci.plugins.github__branch__source.ForkPullRequestDiscoveryTrait' {
-                      strategyId(1)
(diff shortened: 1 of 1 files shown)


timja (approved):

config/default/jenkins-infra.yaml
```diff
@@ -297,235 +297,95 @@ jenkins:
...
-                          ignoreUntrustedChanges(true)
-                        }
-                        buildRegularBranches()
-                        buildTags {
-                          atLeastDays("0")
-                          atMostDays("7")
-                        }
+                    orphanedItemStrategy {
+                      discardOldItems {
+                        daysToKeep(1) // Keep removed SCM heads/branch/PRs only for 1 day
```
olblak, line 349: One day seems a bit short to me, what do you think to put to 7 days?

config/default/jenkins-infra.yaml
```diff
@@ -297,235 +297,95 @@ jenkins:
...
-                organizationFolder('Docker Builds') {
-                  description('Docker Builds')
-                  displayName('Docker Builds')
-                  organizations {
-                    github {
-                      repoOwner("jenkins-infra")
-                      apiUri("https://api.github.com")
-                      credentialsId('github-app-infra')
+                [
+                  ['Docker Builds', 'jenkins-infra', 'docker-.*|^jenkins.io$'],
```
olblak, line 301: Is there any reason to use a list of arrays instead of a list of map? I find the latter one easier to read, but it's not blocker

config/default/jenkins-infra.yaml
```diff
@@ -297,235 +297,95 @@ jenkins:
         jobs-settings: |
           jobs:
             - script: >
```
olblak, line 299: Would it be possible to have a comment that explain the intend of this script and comment the other script as well

olblak: The idea is to make it clearier where to add a git repository

config/default/jenkins-infra.yaml
dduportal: @olblak This setting is only about tracking the pipelines of the merged/closed PR or deleted branches. What would be the goal of keeping it for 7 days? (I would advocate for immediate deletion honestly 😅).
The build log persistence is a different setup, which is done on the pipelines themselves: it's up to the pipeline's writer to define this policy (e.g. no need to change the Jenkins conf for such a senstivie one).

config/default/jenkins-infra.yaml
dduportal: Good amelioration proposal, it would help the reader! 👍

config/default/jenkins-infra.yaml
dduportal: Good point!

dduportal: Done in #1126

config/default/jenkins-infra.yaml
dduportal: Done in #1126

config/default/jenkins-infra.yaml
dduportal: Discussed in 1:1 with @olblak , we keep the 1 day the comment was about the build retention

dduportal: Added comments in #1126  to help the future readers

PR #1126, opened 2021-04-29
Chore: infra.ci JobDSL config refinement

dduportal:
This PR is a follow up of #1124 , to apply @olblak suggestions.

Only syntax refinement: no changes visible

PR #1136, opened 2021-04-30
fix: revert infra.ci config to remove jenkins.io builds

dduportal:
The changes from #1133 were overriden by #1126 . Still not sure how this could have happened (but since we work with squash merged PR, the conflict management behavior can be the culprit here)


MarkEWaite (approved):

MarkEWaite:
Thanks @dduportal!  I've confirmed that changes are flowing through to www.jenkins.io

PR #1218, opened 2021-06-03
Fix: missing security top-level key syntax error on JCasc configuration for infra.ci

dduportal:
The PR #1217 introduced a syntax error, because the key `security` has to be part of the content (not only the name).


dduportal:
I merged this PR using my administrator's privileges after testing that it was working in production, to avoid a rollback of the applied setting from the master branch.

PR #1219, opened 2021-06-03
Remove azure.mirror.jenkins.io from publick8s

olblak:
I would like to remove this mirror from the kubernetes cluster running on Azure.
I am looking for ways to reduce the cost of that account and the network bandwidth is one of the options.

Once this PR is validated, I'll manually remove the service by running `helm uninstall --namespace mirror mirror` and then remove the dns record from Azure dns


dduportal (approved):
Sounds good, let's wait for the CI feedback

PR #1221, opened 2021-06-03
feat: ensure that weather is disabled on infra. and release.ci

dduportal:
Ref. 

- https://github.com/jenkins-infra/docker-jenkins-weekly/pull/204
- https://support.cloudbees.com/hc/en-us/articles/216973327

### DIFF
--- a/config/default/jenkins-infra.yaml
+++ b/config/default/jenkins-infra.yaml
@@ -551,6 +551,10 @@ jenkins:
                   - "Job/Build:authenticated"
                   - "Job/ExtendedRead:authenticated"
         system-settings: |
+          unclassified:
+            defaultFolderConfiguration:
+              # Keep healthMetrics an empty list to ensure weather is disabled
+              healthMetrics: []
           jenkins:
             disabledAdministrativeMonitors:
               - "jenkins.security.QueueItemAuthenticatorMonitor"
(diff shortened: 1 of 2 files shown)


config/default/jenkins-infra.yaml
```diff
@@ -551,6 +551,10 @@ jenkins:
                   - "Job/Build:authenticated"
                   - "Job/ExtendedRead:authenticated"
         system-settings: |
+          unclassified:
+            defaultFolderConfiguration:
+              healthMetrics:
```
timja, line 556: have you tested the above as working? I would've expected empty list to be required, but maybe null is fine?
```suggestion
              healthMetrics: []
```

dduportal: I tested successfully, on a local instance, that: 
- It did not crash (to avoid breaking infra.ci as I did yesterday 😅)
- If I add the default health metric through the UI and reload jenkins, then the settings goes back to empty on the UI (because of Casc).

I understand that JCasc is smart and cast a null value to the "empty" array, but I like explicit thing: let me test and add your suggestion.

Thanks!

dduportal: Validated, sounds good.

timja (approved):

PR #1222, opened 2021-06-03
Update mirrorbits documentation for archives.jenkins.io

olblak:
After some testing with @MarkEWaite , it appears that a score of -1 slightly favor xmission over archives.jenkins.io which is not what we want.
What we want is to be redirected to be redirected to archives.jenkins.io only if no other mirrors are available.


dduportal (approved):
👍 👏

dduportal:
@olblak I'll let you merge this once ready, to be sure you stay on control

PR #1236, opened 2021-06-07
Allow new JFrog IP

olblak:


timja (approved):

olblak:
According this document https://jfrog.com/knowledge-base/what-are-the-artifactory-cloud-nated-ips/ shared by @daniel-beck , we should probably allow every IP from this list as we don't know where our instance is running

timja:
> According this document [jfrog.com/knowledge-base/what-are-the-artifactory-cloud-nated-ips](https://jfrog.com/knowledge-base/what-are-the-artifactory-cloud-nated-ips/) shared by @daniel-beck , we should probably allow every IP from this list as we don't know where our instance is running

our one doesn't appear to be on that list?

olblak:
Nope but I would like to try it anyway, worst case scenario I'll rollblack

PR #1238, opened 2021-06-08
Update JFrog allowed IP

olblak:
Those two IPs were shared from Jfrog and are currently working.
I need confirmation from Jfrog ops team that we don't need other IPs.

Once this PR is merged, we can re-enabled the job on infra.ci


timja (approved):

olblak:
I got confirmation that those two IPs are all what we need

PR #1247, opened 2021-06-11
feat: [jenkins.io] Set header "X-Content-Type-Options" to nosniff [INFRA-2998]

dduportal:
This PR introduces a new Header `X-Content-Type-Options` set to the value `nosniff` for all the ingress rules on the cluster `publick8s` (AKS), including jenkins.io, as per INFRA-2998.

### DIFF
--- a/config/default/jenkinsio.yaml
+++ b/config/default/jenkinsio.yaml
@@ -12,7 +12,7 @@ ingress:
       if ( $scheme = "https" ) {
         more_clear_headers "Strict-Transport-Security";
       }
-
+      more_set_headers "X-Content-Type-Options: nosniff";
   hosts:
     - host: jenkins.io
       paths:


helmfile.d/nginx-ingress.yaml
```diff
@@ -37,6 +37,7 @@ releases:
             # I doubt we need it at the moment, hence this comment.
             # https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-geoip2
             use-geoip2: "true"
+            hsts-max-age: "2592000" # 30 days - INFRA-2998
```
timja, line 40: no reason to use the minimum?

```suggestion
            hsts-max-age: "31536000" # 1 year - INFRA-2998
```

unless you just want to test it first?

config/default/jenkinsio.yaml
```diff
@@ -12,7 +12,7 @@ ingress:
       if ( $scheme = "https" ) {
         more_clear_headers "Strict-Transport-Security";
```
timja, line 13: don't you need to remove this? i.e. revert https://github.com/jenkins-infra/charts/pull/98

as we now access mirrors over https

helmfile.d/nginx-ingress.yaml
dduportal: Yes, I wondered, but it seems (as per https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#configuration-options)  that `15724800` is the default (e.g. 182 days)

config/default/jenkinsio.yaml
dduportal: I'll move the HSTS stuff on another PR

timja (approved):

Wadeck (approved):
No OPS knowledge useful to really review but really happy to see this kind of changes 🎉 That will reduce the number of report we are getting in security ML

PR #1403, opened 2021-08-18
feat: enable HSTS by default on all the publick8s cluster

dduportal:
This PR comes from INFRA-2998 and aims at enabling HSTS globally on the public ingress of the cluster `publick8s` by default, unless disabled per ingress rules.

It reverts the content of https://github.com/jenkins-infra/charts/pull/98 as tipped by @timja , now that we have HTTPS mirrors (and that the charts allows disabling HSTS per ingress rule).

The HSTS values are the default recommended per https://hstspreload.org/.

⚠️ Please note that this PR, once merged and deployed, might break some usages (clictracking, stats, cross sites). It is a non easy topics, so if we breaks usages, we'll have to avoid disabling it and instead fine tune (per website, or header values).
=> It means that a torough reveiw is required

### DIFF
--- a/helmfile.d/nginx-ingress.yaml
+++ b/helmfile.d/nginx-ingress.yaml
@@ -37,6 +37,11 @@ releases:
             # I doubt we need it at the moment, hence this comment.
             # https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-geoip2
             use-geoip2: "true"
+            hsts: "true"
+            hsts-preload: "true"
+            hsts-include-subdomains: "true"
+            # Should be 2 years as per https://hstspreload.org/
+            hsts-max-age: "86400" # 24 hours, as a first gradual HSTS deployment - see INFRA-2998
           replicaCount: 2
           ingressClass: public-ingress
           service:
--- a/charts/uplink/values.yaml
+++ b/charts/uplink/values.yaml
@@ -40,7 +40,6 @@ ingress:
   annotations:
     kubernetes.io/tls-acme: "true"
     kubernetes.io/ingress.class: "nginx"
-    nginx.ingress.kubernetes.io/hsts: "true"
     nginx.ingress.kubernetes.io/rewrite-target: /
     nginx.ingress.kubernetes.io/ssl-redirect: "true"
   hosts:
@@ -71,7 +70,7 @@ tolerations: []
 affinity: {}
 
 postgresql:
-  url: 
+  url:
 client:
   id: ""
   secret: ""
--- a/config/default/accountapp.yaml
+++ b/config/default/accountapp.yaml
@@ -3,7 +3,7 @@ ingress:
   annotations:
     "cert-manager.io/cluster-issuer": "letsencrypt-prod"
     "kubernetes.io/ingress.class": "public-ingress"
-    "nginx.ingress.kubernetes.io/hsts": "true"
+
     "nginx.ingress.kubernetes.io/ssl-redirect": "true"
     "nginx.ingress.kubernetes.io/affinity": "cookie"
 
--- a/config/default/captain-hook.yaml
+++ b/config/default/captain-hook.yaml
@@ -4,7 +4,7 @@ ingress:
   annotations:
     # "cert-manager.io/cluster-issuer": "letsencrypt-prod"
     "kubernetes.io/ingress.class": "public-ingress"
-    "nginx.ingress.kubernetes.io/hsts": "true"
+
     # "nginx.ingress.kubernetes.io/ssl-redirect": "true"
     # tls:
     #   - hosts:
--- a/config/default/custom-distribution-service.yaml
+++ b/config/default/custom-distribution-service.yaml
@@ -3,7 +3,7 @@ ingress:
   annotations:
     "cert-manager.io/cluster-issuer": "letsencrypt-prod"
     "kubernetes.io/ingress.class": "public-ingress"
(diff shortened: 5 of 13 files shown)


MarkEWaite:
My last experience with HSTS enabled on a site was bad due to the "sticky" nature of the HSTS setting.  If I understand correctly, once a browser sees a site is HSTS enabled, there is no way to "undo" that setting.  Agreed that we need careful review and thought before this is merged

dduportal:
That's a good point: we might want to re-introduce HSTS gradually.
WDYT about starting with a 1 day max age for 7 days. If no issue, then we can move to 1 week max age for 2 weeks, and then 1 month for 2 month, until the 2 years limits?

helmfile.d/nginx-ingress.yaml
```diff
@@ -37,6 +37,10 @@ releases:
             # I doubt we need it at the moment, hence this comment.
             # https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-geoip2
             use-geoip2: "true"
+            hsts: "true"
+            hsts-preload: "true"
+            hsts-include-subdomains: "true"
```
olblak, line 42: Enabling HSTS by default on every website under jenkins.io and jenkins-ci.org for two years is seriously annoying. We've been there and decided to only enable it for specific endpoints.
Last time we broke the mirroring infrastructure with that setting which should be fine now we it's available on https but we need to double check

helmfile.d/nginx-ingress.yaml
```diff
@@ -37,6 +37,10 @@ releases:
...
+            hsts-max-age: "63072000" # 2 years as per https://hstspreload.org/
```
olblak, line 43: I agree it's already a nice improvement to increase by default the hsts max age

helmfile.d/nginx-ingress.yaml
olblak: I think that by default I would enable HSTS by default *but* I would not enable it for subdomains so we can still disable HSTS for a specific website if we need it

timja: I _think_ it should be ok, but it will probably break something =/

dduportal:
PR updated with the following changes, trying to take all your feedbacks in account, WDYT?

* Removing the HSTS for subdomains, to ensure any ingress rule can opt-out easily without requiring web browsers to be reset individually
* Defined the max age at 24h for now (with an inline comment), so if we break something, the impact would last only 24h

Please note that there is already an ingress rule that explictly disable HSTS: https://github.com/jenkins-infra/charts/blob/4173be64f6a814eb6b8fd04ddef363e741a5f20c/config/default/mirrorbits.yaml#L26

timja:
> PR updated with the following changes, trying to take all your feedbacks in account, WDYT?
> 
> * Removing the HSTS for subdomains, to ensure any ingress rule can opt-out easily without requiring web browsers to be reset individually
> * Defined the max age at 24h for now (with an inline comment), so if we break something, the impact would last only 24h
> 
> Please note that there is already an ingress rule that explictly disable HSTS:
> 
> https://github.com/jenkins-infra/charts/blob/4173be64f6a814eb6b8fd04ddef363e741a5f20c/config/default/mirrorbits.yaml#L26

HSTS is on most services already though. (including get.jenkins.io, not sure what that ingress rule you mean is doing as it's on)
Can't say I really see the point in this 24h setting if you aren't including sub domains, as it's already enabled.

Setting includeSubDomains with 24h would make sense as it's easier to fix any issues / roll it back if needed

dduportal:
> HSTS is on most services already though. (including get.jenkins.io, not sure what that ingress rule you mean is doing as it's on)
> Can't say I really see the point in this 24h setting if you aren't including sub domains, as it's already enabled.
> 
> Setting includeSubDomains with 24h would make sense as it's easier to fix any issues / roll it back if needed

Thanks @timja! I (wrongly) assumed that the annotation was working, but it is not:

```shell
date && curl -s -vv  https://get.jenkins.io -o /dev/null 2>&1 | grep -i 'x-'
Thu Aug 19 10:38:41 CEST 2021
< strict-transport-security: max-age=15724800; includeSubDomains
```

* https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#hsts => Only through the configmap (e.g. through the `config:` directive for each nginx controller in the helm chart values for us).
* https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/ => does not mention HSTS
* https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/ mentions the annotations `nginx.org/hsts*` but it's with the Nginx Inc. ingress controller (not ours)
* https://github.com/kubernetes/ingress-nginx/issues/549 => HSTS is enabled by default on any TLS `server` block on the ingress-nginx, and the default is 6 month ages (e.g. `15724800` seconds) with subdomains.

The 300s max age from https://www.jenkins.io is not coming from the ingress though: @olblak pointed me that it might be Fastly: that would be a separate configuration update.

Let me update the PR accordingly to your feedbacks

helmfile.d/nginx-ingress.yaml
dduportal: see comments below

dduportal:
PR updated:

- Subdomains are present
- All HSTS annotations removed (because ignored)
- Max Age set to 24h even though it will be overridden by fastly for some domains

timja (approved):

PR #1404, opened 2021-08-18
Increase loki storage

olblak:
The Loki pod is running out of space.
I am not sure yet if it support live resizing


dduportal (approved):
🚀

olblak:
It appears that it can automatically resize the disk, we just have to restart the pod 
https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

PR #1451, opened 2021-09-09
Add allowed IP to ldap service firewall

olblak:
* Allow ldap connections from the new public8s IP range
* Allow inbound connections from the *temporary* confluence backup machine

### DIFF
--- a/config/default/ldap.yaml
+++ b/config/default/ldap.yaml
@@ -17,3 +17,5 @@ service:
     - '52.167.253.43/32'      # Accept inbound LDAPS from public.aks.jenkins.io
     - '34.211.101.61/32'      # Accept inbound connections from Linux Foundation test machine
     - '44.240.22.235/32'      # Accept inbound connections from Linux Foundation prod machine
+    - '20.75.10.224/29'       # Accept inbound connections from publick8s public IPs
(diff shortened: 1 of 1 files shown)


timja (approved):

config/default/ldap.yaml
```diff
@@ -17,3 +17,5 @@ service:
     - '52.167.253.43/32'      # Accept inbound LDAPS from public.aks.jenkins.io
     - '34.211.101.61/32'      # Accept inbound connections from Linux Foundation test machine
     - '44.240.22.235/32'      # Accept inbound connections from Linux Foundation prod machine
+    - '20.75.10.224/29'       # Accept inbound connections from publick8s public IPs
```
MarkEWaite, line 20: Are the ci.jenkins.io agents allocated from the publick82 cluster or are they on a different cluster?  Are they receiving a public IP address in this address range when they are allocated?  Would a malicious program on a ci.jenkins.io agent be able to connect to the LDAP server?
```suggestion
    - '20.75.10.224/29'       # Accept inbound connections from publick8s public IPs
```

MarkEWaite (approved):

config/default/ldap.yaml
olblak: No ci agents are not running on publick8s and should never be allowed

dduportal (approved):
LGTM

PR #1507, opened 2021-09-29
Add keycloak affinity based on cookies

olblak:
* Add connection affinity to avoid authentication being broken when redirected to the second pod
* Increase cpu allocated to keycloak, according to grafana memory doesn't seem to be an issue


dduportal (approved):
Looks good. What is the rational behind the new CPu value?

olblak:
> Looks good. What is the rational behind the new CPu value?

I feel that keycloak is slower than it used to be but I can't find metric to validate my assumption so I am blindly increasing the value

PR #1621, opened 2021-11-02
fix(wiki): use 'wiki-test' as subdomain for final tests and activate TLS [INFRA-3092]

lemeurherve:


dduportal (approved):
LGTM 🚀 

- Code looks good 
- helmfile diff looks good

PR #1625, opened 2021-11-03
chore(wiki): use last version and set final domains [INFRA-3092]

lemeurherve:


config/default/wiki.yaml
```diff
@@ -5,19 +5,19 @@ ingress:
     "cert-manager.io/cluster-issuer": "letsencrypt-prod"
     "nginx.ingress.kubernetes.io/ssl-redirect": "true"
   hosts:
-    - host: wiki-test.jenkins.io
```
timja, line 8: would be good to have some home page content before making this live I think:

https://wiki-test.jenkins.io

also the 404 page is referring to freenode irc

dduportal (dismissed):
LGTM 👍 

Please note that merging this change will make the certificate-manager to fail. I assume it will be available with an insecure certificate until the DNS for `wiki.*` are updated.

config/default/wiki.yaml
lemeurherve: > would be good to have some home page content before making this live I think:

Agreed.

> also the 404 page is referring to freenode irc

Not sure I follow, do you have an example?

timja: Previously the home page was returning a 404 and there's a butler page that says to contact jenkins-infra on freenode IRC

timja (approved):

PR #1658, opened 2021-11-08
chore(cleanup) ensure that helmfile configuration are located in the correct directory named after their cluster (INFRA-3081)

dduportal:
This PR should NOT break the infra but a carefull review of the helmfile diff is required


lemeurherve (approved):
👍

PR #1685, opened 2021-11-10
feat: use official jenkins chart instead of jenkins-infra/jenkins

lemeurherve:


lemeurherve:
Note to reviewer(s): tick "hide whitespaces"
image

lemeurherve:
I'm searching how to either:
- retrieve [the name of the secret](https://github.com/jenkinsci/helm-charts/blob/25df294fed4fa85a4197df9a12e60a7a287ca59f/charts/jenkins/templates/secret.yaml#L6) generated from jenkins chart, to inject it in [the controller values](https://github.com/jenkins-infra/public-charts/blob/9af083f2646e01ed22eb935cd3ab4b4144099c81/charts/jenkins/values.yaml#L13)
- set [the name of the secret](https://github.com/jenkinsci/helm-charts/blob/25df294fed4fa85a4197df9a12e60a7a287ca59f/charts/jenkins/templates/secret.yaml#L6) (should I open a PR on jenkinsci/helm-chart to allow specifying a custom name for the secret?)

dduportal:
Do you think that https://plugins.jenkins.io/kubernetes-credentials-provider/ could help to avoid decoupling the secret's lifecycle from the jenkins-release release?

lemeurherve:
> Do you think that https://plugins.jenkins.io/kubernetes-credentials-provider/ could help to avoid decoupling the secret's lifecycle from the jenkins-release release?

Maybe? Idk, I'll have to look into it. Probably as a further step.
I think I'll generate them (and the CRB) with the help of a (new) chart, like env-release

dduportal:
> > Do you think that https://plugins.jenkins.io/kubernetes-credentials-provider/ could help to avoid decoupling the secret's lifecycle from the jenkins-release release?
> 
> Maybe? Idk, I'll have to look into it. Probably as a further step. I think I'll generate them (and the CRB) with the help of a (new) chart, like env-release, to keep the same process as the current one, at least for now.

Yep, I'll check to see if I can help here, because on short term we might want to migrate "as it". Thanks for starting this huge work!

lemeurherve:
With the new jenkins-additional chart, and by temporary manually modifying the existing secret in publick8s cluster swapping `meta.helm.sh/release-name: jenkins-infra` annotation with `meta.helm.sh/release-name: jenkins-additional` to avoid the "invalid ownership metadata" `helmfile diff error` , I've come to this diff:

image

lemeurherve:
I don't see anything else problematic in the diff.

I think manually swapping its annotation (or renaming it to keep a backup) should be enought/not too risky to do in order to merge this PR.

WDYT @dduportal ?

dduportal:
> I don't see anything else problematic in the diff.
> 
> I think manually swapping its annotation (or renaming it to keep a backup) should be enought/not too risky to do in order to merge this PR.
> 
> WDYT @dduportal ?

That would work nicely. We only have to write a message in the #jenkins-infra channel to let other know that infra.ci and release.ci will be worked on with outages.

dduportal:
- The diff looks good to me 👏 (only comments or checksums changed)

dduportal (approved):
LGTM great work!

PR #1693, opened 2021-11-12
fix(jenkins-infra) correct cluster-admin role to ensure it exists for…

dduportal:
… the charts job

### DIFF
--- a/helmfile.d/jenkins-infra.yaml
+++ b/helmfile.d/jenkins-infra.yaml
@@ -8,14 +8,12 @@ releases:
     chart: jenkins/jenkins
     version: 3.8.8
     namespace: jenkins-infra
-    set:
-      - name: namespace
-        value: jenkins-infra
     values:
     - ../config/jenkins-infra.yaml
   - name: jenkins-infra-additional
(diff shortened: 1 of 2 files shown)


helmfile.d/jenkins-infra.yaml
```diff
@@ -8,14 +8,12 @@ releases:
...
     - ../config/jenkins-infra.yaml
   - name: jenkins-infra-additional
     chart: jenkins-infra/jenkins-additional
     version: 0.0.1
     namespace: jenkins-infra
-    values:
-    - ../config/jenkins-infra-additional.yaml
+    set:
+      - name: clusterAdminEnabled
+        value: true
```
lemeurherve, line 19:
```suggestion
    values:
      - clusterAdminEnabled: true
```
I would have kept the separate values file, even if it contains only one value, in order to be coherent with the other helmfiles, but specify the only value like that is acceptable.

dduportal:
Sorry, I forgot to close this PR. It was used to debug the jenkins-infra issue with the cluster.

PR #1802, opened 2021-12-14
fix: scaledown keycloak from 2 to 1 replica

lemeurherve:
We're scaling down Keycloak because the session sharing isn't working as intended.


dduportal (approved):
🚀

timja (approved):

PR #1809, opened 2021-12-15
fix: use the new matrix permissions syntax introduced with matrix-auth v3.0

lemeurherve:


config/jenkins-release.yaml
```diff
@@ -479,10 +479,12 @@ controller:
           authorizationStrategy:
             globalMatrix:
               permissions:
-                - "Overall/Administer:release-core"
-                - "Overall/SystemRead:authenticated"
-                - "Overall/Read:authenticated"
-                - "Job/Read:authenticated"
+                - "GROUP:Overall/Administer:release-core"
+                - "GROUP:Overall/Administer:admins"
```
lemeurherve, line 483: @daniel-beck @timja in common with @dduportal we've added this permission for our "admin" group on release.ci.jenkins.io, as we have already admin rights on other services.
OK for you?

config/jenkins-release.yaml
```diff
@@ -479,10 +479,12 @@ controller:
...
+                - "GROUP:Overall/Administer:jenkins-admins"
```
lemeurherve, line 484: @daniel-beck @timja again in common with @dduportal we were wondering if the "jenkins-admin" group should be added to release.ci.jenkins.io.
WDYT?

config/jenkins-release.yaml
daniel-beck: This isn't cert.ci so up to infra team 🤷

timja (approved):
fine with me, the people are generally the same anyway, tbh the general release team doesn't really need admin anyway just build.

dduportal (approved):
👍

PR #1881, opened 2022-01-03
fix: use unique complete image to avoid calling kubectl exec and getting (tcp) errors

lemeurherve:
Using a custom temp image, should finally be the result of [this PR](https://github.com/jenkins-infra/docker-helmfile/pull/45).


dduportal (approved):
🚀

PR #1890, opened 2022-01-04
chore: add an 'ext_' prefix to external charts config files

lemeurherve:


dduportal (approved):
🚀 

The errors come from the updatecli 's behavior with SCM/local SCM.

config/NOTES.txt
```diff
@@ -0,0 +1,4 @@
+External charts config files are prefixed with "ext_" to be easily noticed.
+
+In these config, we're setting/surcharging and updating the image.tag with udpatecli (See ./updatecli/updatecli.d/docker-images).
```
timja, line 3: s/udpatecli/updatecli

lemeurherve: Grrr... I have to find a tool/trick to avoid this error which bitten me more than once already...

halkeye: https://pre-commit.com/index.html with codespell plugin maybe?

lemeurherve: I've got the codespell plugin already, reading the pre-commit link, thanks for the suggestion @halkeye :)

PR #1907, opened 2022-01-12
Allow untrusted builds for jenkins.io and plugin site (which have safe pipelines)

halkeye:
🤞🤞


halkeye:
Plugin site is already updated, but https://github.com/jenkins-infra/jenkins.io/pull/4814 needs to be merged for jenkins.io to be safe

dduportal (changes requested):
Let's wait for the security release :)

dduportal (approved):
🚀

PR #1993, opened 2022-02-09
Updatecli ami correction

smerle33:
trying to add infos to the regexp to correct behavior


lemeurherve (approved):

updatecli/updatecli.d/charts/jenkins-agents-infra-release.yaml
```diff
@@ -69,17 +69,17 @@ targets:
...
   setUbuntuAgentAMIArm64:
     name: "Bump AMI ID for ARM 64"
     kind: file
     sourceID: getLatestUbuntuAgentAMIArm64
     spec:
       file: config/ext_jenkins-infra.yaml
-      matchPattern: '- ami: ".*"((.|\n)*)description: "ARM64'
-      replacePattern: '- ami: "{{ source `getLatestUbuntuAgentAMIArm64` }}"${1}description: "ARM64'
+      matchPattern: '- ami: ".*"((.|\n)*)description: "ARM64 Ubuntu 20.04"'
+      replacePattern: '- ami: "{{ source `getLatestUbuntuAgentAMIArm64` }}"${1}description: "ARM64 Ubuntu 20.04"'
```
dduportal, line 82: Not sure why but this one still tries to modify the AMD64:

```diff
setUbuntuAgentAMIArm64
----------------------
**Dry Run enabled**
⚠ File content for "/tmp/updatecli/jenkins-infra/kubernetes-management/config/ext_jenkins-infra.yaml", updated. 
--- /tmp/updatecli/jenkins-infra/kubernetes-management/config/ext_jenkins-infra.yaml
+++ /tmp/updatecli/jenkins-infra/kubernetes-management/config/ext_jenkins-infra.yaml
@@ -327,7 +327,7 @@
                 sshKeysCredentialsId: "ec2-agents-privkey"
                 useInstanceProfileForCredentials: false
                 templates:
-                - ami: "ami-0320d578c21e1255c" # https://github.com/jenkins-infra/packer-images/
+                - ami: "ami-014ba4a5098e5e684" # https://github.com/jenkins-infra/packer-images/
                   amiOwners: "200564066411"
                   amiType:
                     unixData:
```

https://infra.ci.jenkins.io/blue/organizations/jenkins/kubernetes-management/detail/PR-1993/2/pipeline/37#step-38-log-1387

Damn this non-parseable yaml. We might want to find a way to either
- move the JCasc snbippets into their own "yaml" files
- or use variables/values to specify the ami values

dduportal: Do you think that we could use the comment as a matching pattern (and change to comment to have a string like `windows-amd64, linux-arm64, etc.) ?

I know it's fragile, but if the comment explains why, that could allows a quick win here

dduportal (changes requested):

updatecli/updatecli.d/charts/jenkins-agents-infra-release.yaml
smerle33: we don't have enough lines to be sure but : 0320d578c21e1255c is a windows VM and i am wondering if the previous runs have not screw the good id and that now it tries to put the good one ....

dduportal:
- Let's go for the "comment" solution as a quick fix.
- Do not forget to rebase from `upstream/main` + ensure that you revert the change on #1980 which disabled the manifest.

PR #2058, opened 2022-03-06
chore(updatecli) cleanup updatecli manifest to fix errors and warnings

dduportal:
- 1 fixed error comes from a duplicated value in `updatecli/values.yaml` from https://github.com/jenkins-infra/kubernetes-management/pull/2051. Ping @lemeurherve can you check that I've set the correct value for `github.user`?
- 1 fixed error comes from a `pullrequest` block not specifying the correct `target`
- Warnings around Docker image missing architecture

### DIFF
--- a/updatecli/values.yaml
+++ b/updatecli/values.yaml
@@ -1,5 +1,5 @@
 github:
-  user: "updatebot"
+  user: "Jenkins Infra Bot (updatecli)"
   email: "60776566+jenkins-infra-bot@users.noreply.github.com"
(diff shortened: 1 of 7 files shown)


updatecli/values.yaml
```diff
@@ -1,6 +1,5 @@
 github:
   user: "updatebot"
```
lemeurherve, line 2:
```suggestion
  user: "Jenkins Infra Bot (updatecli)"
```

lemeurherve (changes requested):

lemeurherve (approved):

PR #2103, opened 2022-03-17
fix: suppress log output in github checks

lemeurherve:
Avoid secret leaks when build fails

### DIFF
--- a/config/ext_jenkins-infra.yaml
+++ b/config/ext_jenkins-infra.yaml
@@ -491,6 +491,17 @@ controller:
                       credentialsId('github-app-infra')
 
                       traits {
+                        gitHubSCMSourceStatusChecksTrait {
+                            // Note: changing this name might have impact on github branch protections if they specify status names
+                            name('[infra.ci.jenkins.io] ' + config.name)
+                            skip(false)
+                            // If this option is checked, the notifications sent by the GitHub Branch Source Plugin will be disabled.
+                            skipNotifications(false)
+                            skipProgressUpdates(false)
+                            // Default value: false. Warning: risk of secret leak in console if the build fails
+                            suppressLogs(true)
+                            unstableBuildNeutral(false)
+                        }
                         gitHubTagDiscovery()
                         cloneOptionTrait {
                           extension {


dduportal (approved):
LGTM (tested in live)

config/ext_jenkins-infra.yaml
```diff
@@ -491,6 +491,17 @@ controller:
                       credentialsId('github-app-infra')
 
                       traits {
+                        gitHubSCMSourceStatusChecksTrait {
+                            // Note: changing this name might have impact on github branch protections if they specify status names
+                            name('[infra.ci.jenkins.io] ' + config.name)
+                            skip(false)
+                            // If this option is checked, the notifications sent by the GitHub Branch Source Plugin will be disabled.
+                            skipNotifications(false)
```
timja, line 499: isn't this what we want?
```suggestion
                            skipNotifications(true)
```

config/ext_jenkins-infra.yaml
```diff
@@ -491,6 +491,17 @@ controller:
                       credentialsId('github-app-infra')
 
                       traits {
+                        gitHubSCMSourceStatusChecksTrait {
+                            // Note: changing this name might have impact on github branch protections if they specify status names
+                            name('[infra.ci.jenkins.io] ' + config.name)
```
timja, line 496: do we need this? what does it look like rendered?

dduportal: Looks like this:

Capture d’écran 2022-03-17 à 18 58 44

We found it pretty useful to immediately see which controller is used.

config/ext_jenkins-infra.yaml
dduportal: We understood that the notification is the status of the build (started, failed, succeeded) to have the red/yellow/green icon so we stil lwant that.


It's the `suppressLogs(true)` that we want so that we keep a feedback (to help public contributors) with only the stage name / sh command / exit status.

Note that we also tried with `skip(true)` : in that case the pipeline exit status is still reported in the PR conversation tab, with the red/yellow/green icon, and the github checks is set to zero with a direct link to the Jenkins build. Safer but less easy for external contributor.

WDYT?

timja: ah right i misread

PR #2107, opened 2022-03-18
feat(infra.ci) status check names

lemeurherve:
Also, suppress log output in github checks to avoid secret leaks when build fails


dduportal (approved):
🚀

PR #2112, opened 2022-03-21
feat(cik8s,doks): add dockerhub image pull secret for datadog and jenkins-agents

smerle33:
as per jenkins-infra/helpdesk#2830 
secret added in : https://github.com/jenkins-infra/charts-secrets/commit/f44237af5ee0ffb40e7998d7f0dc07e1d17246d6


clusters/ext_docker-registry-secrets.yaml
```diff
@@ -0,0 +1,5 @@
+imageCredentials:
+  enabled: true
+  namespaces:
```
lemeurherve, line 3:
```suggestion
  namespaces:
```

clusters/ext_docker-registry-secrets.yaml
```diff
@@ -0,0 +1,5 @@
+imageCredentials:
```
lemeurherve, line 1: This file should be renamed `docker-registry-secrets.yaml`, cf comment above.

lemeurherve (changes requested):

timja (approved):

clusters/cik8s.yaml
```diff
@@ -13,7 +13,17 @@ repositories:
   - name: jenkins-infra
     url: https://jenkins-infra.github.io/helm-charts
 releases:
+  - name: docker-registry-secrets
+    namespace: default
```
lemeurherve, line 17: I wouldn't put it in the "default" namespace, maybe "docker-registry" here?

smerle33: agreed, but @dduportal said to go by step ;) ... 
still a WIP

dduportal: We discovered that specifying a nemspace for the helmfile release is required (despite what the doc. says) to allow referencing from other release (with the `needs` keyword).

Since this helm chart does not create any resource in the aformentioned namespace (it creates in the value-specified namespaceS), I proposed to use "defaukt" here, as a "no-op" value.

Stephane just added a comment in the code to explain this. WDYT?

dduportal (approved):

clusters/cik8s.yaml
lemeurherve: I dunno about the "default" namespace, which is frenquently used for tests or install without being cautious, and subject to deleting. Not a big problem per se, but I'd put it in "jenkins-agents", WDYT?

PR #2179, opened 2022-04-05
chore: add cluster name in last stage name

lemeurherve:
As the "matrix" label is cut when displayed in Blue Ocean

Replaces https://github.com/jenkins-infra/kubernetes-management/pull/2178 (shorter label)


lemeurherve:
Seems to be not possible that way: https://issues.jenkins.io/browse/JENKINS-61280

PR #2198, opened 2022-04-08
feat: GHA to delete unmerged branch

lemeurherve:
From https://github.com/asaaki/gha-delete-unmerged-branches

### DIFF
--- a/.github/workflows/delete-unmerged-branch.yaml
+++ b/.github/workflows/delete-unmerged-branch.yaml
@@ -1,0 +1,26 @@
+name: Delete unmerged branch
+
+on:
+  pull_request:
+    branches: [ main ]
+    types: [ closed ]
+
+jobs:
+  worker:
+    if: ${{ github.event.pull_request.merged_at == '' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: delete-unmerged-branch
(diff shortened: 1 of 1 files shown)


dduportal:
🤔 looks good however I wonder:

- Do we have to add this workflow on each of our repos? Or would it be possible to have it executed at organization level?
- Given how it works, do you think that it could be a Jenkins pipeline so we avoid GHA as much as posible (now that we got gh app token)?

lemeurherve:
>     * Do we have to add this workflow on each of our repos? Or would it be possible to have it executed at organization level?

We can try :)

>     * Given how it works, do you think that it could be a Jenkins pipeline so we avoid GHA as much as posible (now that we got gh app token)?

I'm not sure how I could trigger it on "closed PR" events, but if it's possible why not.
Wondering where I could put this so it's running on all our repos though.

halkeye:
> I'm not sure how I could trigger it on "closed PR" events, but if it's possible why not.

I wrote up notes somewhere on an extension to pipeline to run pipelines for disabled/deleted jobs, but afaik thats not possible as is.

Could make a new utils type repo that just runs scripts on gh or something.

timja (approved):

PR #2201, opened 2022-04-09
feat: Update presentation embed

NotMyFault:
The frame I suggested to put the design library topic in looks fine on Jenkins' default theme, but dark theme or solarized don't look quite well when working with static colors.

### DIFF
--- a/config/ext_jenkins-weekly.yaml
+++ b/config/ext_jenkins-weekly.yaml
@@ -19,7 +19,7 @@ networkPolicy:
       name: "jenkins-weekly"
 primaryView:
   all:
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

NotMyFault:
Is there something I need to do on my side to make the check run? It looks like auto merging is stalled, because it doesn't run at all 👀

halkeye:
right, these jobs are setup to only run when someone with write access commits. Lemme see if i can kick it off manually.

halkeye:
muhaha, thats a yes :P

NotMyFault:
Thanks for taking care of it, Gavin 👏🏻

PR #2137, opened 2022-03-30
feat(rating): new release for rating

smerle33:
Related to https://github.com/jenkins-infra/helpdesk/issues/1627 

add a release for rating with corresponding secrets.yaml

PR #2329, opened 2022-05-05
feat(mirrorbits) add ingress for mirrors.jenkins.io - jenkins-infra/helpdesk#2888

dduportal:
This change request is related to jenkins-infra/helpdesk#2888.

It adds another ingress rule for the `mirrorbits` service, along with the current `get.jenkins.io`.

Please note that:

- TLS certificates will be generated without any problem, like it was done for rating.jenkins.io in #2137 : cert manager uses an ACME DNS resolver.
- Once deployed, This change will be testable by changing your local `/Etc


timja (approved):

dduportal:
Woops I did a typo on the domain name. Fixed (was missing the `s`of mirror)

PR #2399, opened 2022-05-24
fix(jenkinsistheway redirection): use new chart and add dependencie

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/2943
new chart with ingressClassName 
add updatecli yaml to bump the chart version


dduportal (approved):
- Fix for the ingressclass is picked by the `helmfile diff` ✅ 
- Updatecli addon works as expected (as per the build logs) ✅ 
- The "needs" works as expected ✅

PR #2473, opened 2022-06-07
fix: revert jenkinsUrl

lemeurherve:
Revert the change introduced in https://github.com/jenkins-infra/kubernetes-management/pull/2468 which prevented agent to connect to the controller on release.ci.jenkins.io (cf https://release.ci.jenkins.io/job/core/job/weekly/job/release/job/master/135/console)

### DIFF
--- a/config/ext_jenkins-release.yaml
+++ b/config/ext_jenkins-release.yaml
@@ -191,7 +191,7 @@ controller:
           clouds:
             - kubernetes:
                 containerCapStr: "100"
-                jenkinsUrl: "http://default-release-jenkins.svc.cluster.local:8080"
+                jenkinsUrl: "http://default-release-jenkins:8080"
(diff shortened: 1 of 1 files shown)


dduportal:
For context, we applied the same change as in #2468 on infra.ci in https://github.com/jenkins-infra/kubernetes-management/pull/2434, because ot was required to allow Jenkins to be reachable by the windows pod agents.

It appears that it is not working on release.ci.jenkins.io.

As these 2 controller are hosted on 2 distincts AKS clusters (`temp-privatek8s` for infra.ci and `prodpublick8s` for release.ci), I assume that the internal cluster DNS might be differents (different CoreDNS versions or configuration).

smerle33 (approved):
sounds good 🚀

dduportal:
Merging without waiting for the build, to avoid any risk.

PR #2536, opened 2022-06-23
fix(weekly.ci): correct place to store the description

lemeurherve:
Testing different combinations, we determined the `primaryView` wasn't in the correct section of the Jenkins helm chart values.yaml file (not in the `jcasc` section), but also its value wasn't sufficient to be taken in account.
Even more, it's the `views.-all` parameter (and only this one) which has to be set so the text could be displayed.
By setting this parameter, when we look at the resulting JCasc export, we can see the `primaryView` has been set too, with the same value:

Setting both values isn't working in the sense the `primaryView` submitted value will be overwritten by the `view.-all` one.
(I don't understand the goal of this, or why the `primaryView` isn't just a reference to the `views.-all`)

Fixes https://github.com/jenkins-infra/helpdesk/issues/2999

### DIFF
--- a/config/ext_jenkins-weekly.yaml
+++ b/config/ext_jenkins-weekly.yaml
@@ -17,10 +17,6 @@ networkPolicy:
     allowed: true
     namespaceLabels:
       name: "jenkins-weekly"
-primaryView:
-  all:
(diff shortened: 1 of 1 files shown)


timja (approved):
if it works

lemeurherve:
> if it works

Manually determined and tested by editing config map on the cluster where weekly.ci is running.

PR #2571, opened 2022-07-06
hotfix(javadoc/jenkins.io/plugin-site/reports): specify secretNamespace

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3037

### DIFF
--- a/config/javadoc.yaml
+++ b/config/javadoc.yaml
@@ -27,5 +27,6 @@ resources:
 htmlVolume:
   azureFile:
     secretName: javadoc
+    secretNamespace: javadoc
     shareName: javadoc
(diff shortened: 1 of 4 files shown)


lemeurherve:
There is no need for this PR as these charts use [azureFile volumes](https://kubernetes.io/docs/concepts/storage/volumes/#azurefile), without PV/PVC.

Note this type of volume is now deprecated, we have to change these to use Azure CSI.

dduportal:
=> Confirmed by the PR that fixed this bug in Kubernetes: https://github.com/kubernetes/kubernetes/pull/108000/files it is scoped to persistent volumes

PR #2597, opened 2022-07-13
fix: create DOKS auth token with a small life duration

lemeurherve:
Fix the similar issue we've got: https://github.com/digitalocean/action-doctl/issues/30


dduportal (approved):

lemeurherve:
Won't do, only dotcl kubectl show and save commands have this option, and we don't care that much having a lot of token expiring 7 days later.

PR #2646, opened 2022-07-26
chore: test new docker image from packer-images

lemeurherve:


dduportal (approved):
Not tested, but LGTM at first glance

Issue #1668, opened 2021-11-09
Manage account app's docker image tag


dduportal:
Requires https://github.com/jenkins-infra/account-app/issues/134 to be implemented

- [x] Update the account app's helmfile values to specify the tag image (and pull policy to ifnotpresent)
- [x] Add an updatecli manifest to track the docker image tag
- [x] Migrate `Jenkinsfile` to `Jenkinsfile_k8s`
- [x] Remove job from trusted.ci.jenkins.io

dduportal:
Next step before closing: add updatecli automatic tracking of the image tag

PR #2717, opened 2022-08-17
Bump account-app image to 0.0.1 (new tag scheme)

dduportal:
The goal is to deploy the change https://github.com/jenkins-infra/account-app/pull/136#event-7205302072 to production.

We haven't deployed this app since ages so let's keep an eye.

Note: it benefits from https://github.com/jenkins-infra/kubernetes-management/issues/1668

cc @daniel-beck @wadeck


lemeurherve (approved):

dduportal:
helmfile diff looks good: 🚢

PR #2721, opened 2022-08-18
fix: use ec2-plugin 'javaPath' option instead of command prefix

lemeurherve:
Should fix Windows (which can't have command prefix)

Ref:
- jenkinsci/ec2-plugin#766
- jenkins-infra/helpdesk#3090
- jenkins-infra/docker-jenkins-weekly#565

### DIFF
--- a/config/ext_jenkins-infra.yaml
+++ b/config/ext_jenkins-infra.yaml
@@ -306,7 +306,7 @@ controller:
                   hostKeyVerificationStrategy: ACCEPT_NEW     # TODO: set it CHECK_NEW_HARD (slower but safer)
                   idleTerminationMinutes: "30"                # Windows instances are billed per hour, let's reuse
                   instanceCapStr: "50"
-                  javaPath: "C:/tools/jdk-11/bin/java"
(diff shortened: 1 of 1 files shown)


config/ext_jenkins-infra.yaml
```diff
@@ -308,6 +306,7 @@ controller:
                   hostKeyVerificationStrategy: ACCEPT_NEW     # TODO: set it CHECK_NEW_HARD (slower but safer)
                   idleTerminationMinutes: "30"                # Windows instances are billed per hour, let's reuse
                   instanceCapStr: "50"
+                  javaPath: "C:/tools/jdk-11/bin/java"
```
lemeurherve, line 309:
```suggestion
                  javaPath: "C:/tools/jdk-11/bin/java.exe"
```

config/ext_jenkins-infra.yaml
```diff
@@ -223,7 +223,6 @@ controller:
                   amiType:
                     unixData:
                       sshPort: "22"
```
lemeurherve, line 225: "22" is already the default sshPort value, we can delete each of these blocs.

```suggestion
```

PR #2748, opened 2022-08-25
fix(nginx-proxy): set resources to avoid a qos 'bestEffort' and add basic-auth

lemeurherve:


config/nginx-proxy-common.yaml
```diff
@@ -0,0 +1,11 @@
+# Common values for all nginx-proxies
+resources:
+  limits:
+    cpu: 100m
+    memory: 128Mi
+  requests:
+    cpu: 100m
+    memory: 128Mi
```
dduportal, line 8: Good for starting, but it seems a bit low for the CPU. Take care of checking the datadog metrics (CPU load especially)

dduportal (approved):

PR #2761, opened 2022-08-26
chore: bump nginx-proxy

lemeurherve:
Should fix:
> Error: release nginx-proxy-azure failed, and has been uninstalled due to atomic being set: PersistentVolume 'nginx-proxy-azure-pv' is invalid: spec: Required value: must specify a volume type


smerle33 (commented):
🚀

PR #2786, opened 2022-09-01
feat: build only modified cluster(s) if there isn't any config change

lemeurherve:
If there is only a cluster definition change: build only this cluster
If there is any (chart) `config` change: build on all clusters as we can't determine easily and reliably if a specific cluster is concerned by the (chart) change
If there is a cluster definition change and any (chart) `config` change: same as above, build on all clusters

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -90,6 +90,12 @@ pipeline {
       }
         stages {
           stage('Prepare Environment'){
+            when {
+              anyOf {
+                  changeset "clusters/${K8S_CLUSTER}.yaml"
+                  changeset "config/**"
+              }
+            }
             steps {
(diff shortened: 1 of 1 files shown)


Jenkinsfile_k8s
```diff
@@ -90,6 +90,12 @@ pipeline {
       }
         stages {
           stage('Prepare Environment'){
+            when {
+              anyOf {
+                  changeset "clusters/${K8S_CLUSTER}.yaml"
+                  changeset "config/**"
```
lemeurherve, line 96: Wondering if I shouldn't add "Jenkinsfile_k8s" to these paths (?)
Can be added later if needed.

smerle33 (approved):
🚀

PR #2817, opened 2022-09-02
feat: separate updatecli stages to their own pipeline

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2778


config/ext_jenkins-infra-jobs.yaml
```diff
@@ -121,6 +121,25 @@ jobsDefinition:
             fileName: "kubeconfig"
             description: "Kubeconfig file for temp-privatek8s"
             secretBytes: "${base64:${KUBECONFIG_TEMP_PRIVATEK8S}}"
+      kubernetes-management-updatecli:
+        name: Kubernetes Management Dependencies
+        repository: kubernetes-management
```
lemeurherve, line 126: @dduportal At first I've wondered how I could specify a second job on the same repository, looked at your jenkins-jobs chart and found how the id or the .repository value of the job are coalesced. Again, really nice job! 👏 (pun intended)

dduportal (commented):
https://github.com/jenkins-infra/helpdesk/issues/2778 specified a top-level Job of type "GitHub organization scanning" that would search for each `Jenkinsfile_updatecli` for each branch of each repository.

I understand that you are creating a job alongside the current "kubernetes-management": is there a particular reason for this choice? 

note that I'm not against your proposal, neither in favor, it's a neutral question to be sure that I can catch up.

Also, my question is motivated because since we changed from (gh-org job) to (folder of multibranches jobs), there might be a lot of good reasons

lemeurherve:
Changed to a top level folder job with one children for now.

We'll adapt the jenkins-jobs chart to allow specifying a job of kind: folder with specific jenkinsfilePath, see https://github.com/jenkins-infra/helm-charts/issues/105

PR #2830, opened 2022-09-07
fix: remove datadog release specific timeout

lemeurherve:
Was put in place for a previous problem. This timeout causes troubles now as datadog release is frenquently in a bad state:
'Error: UPGRADE FAILED: another operation (install/upgrade/rollback) is in progress'


dduportal (approved):
❤️ 
🚢

PR #2843, opened 2022-09-08
feat(datadog) enable HA mode for cluster agent + enable network monitoring

dduportal:
The logs of the cluster-agent components recommens to run in HA (High Availability) mode. This PR enable the HA mode by setting the amount of replicas to `2` as per https://github.com/DataDog/helm-charts/tree/main/charts/datadog.

It's also the opportunity to enable the network advanced monitoring by datadog (https://github.com/DataDog/helm-charts/tree/main/charts/datadog#enabling-npm-collection).

Please note that we used to set up an incorrect value for the cluster agent: it must be singular form. It was "partially" working because the default value is to have it enabled since months. By fixing this "syntax" error, it means that more datadog components will be created in the helmfile diff.

### DIFF
--- a/config/ext_datadog.yaml.gotmpl
+++ b/config/ext_datadog.yaml.gotmpl
@@ -38,9 +38,10 @@ agents:
     rollingUpdate:
       maxUnavailable: "33%"
     type: RollingUpdate
-clusterAgents:
-  enabled: true
+clusterAgent:
   metricsProvider:
     enabled: true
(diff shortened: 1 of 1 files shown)


dduportal:
Helmfile diff is what is expected: for each of the 4 clusters, it adds the custom datadog metrics service/endpoints, the sysprobe used for the network monitoring and scales the replicas. Let's ship it

PR #2851, opened 2022-09-09
chore: conditional helmfile commands on pull request

lemeurherve:
Lint and diff on pull requests only if the current cluster helmfile or any of the chart configurations files have changed

Revert https://github.com/jenkins-infra/kubernetes-management/pull/2790

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -139,9 +139,6 @@ pipeline {
                 }
               }
             }
-            when {
-              changeRequest()
-            }
             steps {
               script {
                 def diff = sh(
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -100,11 +100,45 @@ pipeline {
             }
           }
           stage('Helmfile Lint'){
+            // Always lint on main branch
+            // Lint on pull requests only if the current cluster helmfile or any of the chart configurations files have changed
+            when {
+              anyOf {
+                branch 'main'
+                allOf {
+                  changeRequest()
+                  anyOf {
+                    changeset "config/**"
+                    expression {
+                      sh(returnStatus: true, script: '''
+                        git status -s | grep "^clusters/${K8S_CLUSTER}.yaml"
+                        '''
+                      ) == 0
+                    }
+                  }
+                }
+              }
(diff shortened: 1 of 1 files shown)


Jenkinsfile_k8s
```diff
@@ -100,13 +100,44 @@ pipeline {
...
+            // Diff on pull requests only if the current cluster helmfile or any of the chart configurations files have changed
             when {
-              changeRequest()
+              allOf {
+                changeRequest()
+                anyOf {
+                  changeset "config/**"
+                  expression {
+                    sh(returnStatus: true, script: '''
+                      git status -s | grep "^clusters/${K8S_CLUSTER}.yaml"
```
halkeye, line 135: doesn't changeset let you do that?

halkeye: `when { changeset pattern: "^clusters/${K8S_CLUSTER}.yaml", comparator: "REGEXP" }`

lemeurherve: I've tried before with `changeset "clusters/${K8S_CLUSTER}.yaml" but it didn't worked. (Tested again with a replay here: https://infra.ci.jenkins.io/job/kubernetes-jobs/job/kubernetes-management/job/PR-2852/3/console)

I've just tried your syntax in another replay but it doesn't seem to work either: https://infra.ci.jenkins.io/job/kubernetes-jobs/job/kubernetes-management/job/PR-2852/4/console

Unfortunately, git+grep doesn't seem to work too: https://infra.ci.jenkins.io/job/kubernetes-jobs/job/kubernetes-management/job/PR-2852/5/console

halkeye: well grep by default isn't using regex, so you want egrep or grep -e

I'll try to take a look at those links when i'm on a computer i can run vpn again.

dduportal:
I wonder if the (code) complexity / value is still interesting: it adds a lot of (untested -because it's a pipeline) code paths without speeding up the overall result (the 4 clusters are diff-ed in parallel).

What improvements did you see in your experiments?

lemeurherve:
> I wonder if the (code) complexity / value is still interesting

Probably not, closing.

PR #2866, opened 2022-09-12
feat: separate updatecli to its own pipeline

lemeurherve:
Ref: jenkins-infra/helpdesk#2778

(Reopening https://github.com/jenkins-infra/kubernetes-management/pull/2817 after https://github.com/jenkins-infra/kubernetes-management/pull/2865)


updatecli/updatecli.d/pod-templates/helmfile.yaml
```diff
@@ -54,7 +54,7 @@ targets:
     name: Update docker-helmfile in Jenkinsfile
     kind: file
     spec:
-      file: Jenkinsfile_k8s
+      file: Jenkinsfile_updatecli
```
dduportal, line 57:
```suggestion
      files: 
      - Jenkinsfile_k8s
      - Jenkinsfile_updatecli
```

Don't we need the update on both pipelines?

lemeurherve: No, the `jenkinsciinfra/helmfile` image is defined only in the updatecli stages, no mention of it in the remaining Jenkinsfile_k8s

lemeurherve: As a matter of fact, this image shouldn't be defined in either of them as it's already the default included image specified and tracked in the shared pipeline library:
https://github.com/jenkins-infra/pipeline-library/blob/master/vars/updatecli.groovy#L10

One less dependencies location to deal with! Incoming commit.

dduportal: You are correct: I missed the fact that it's centralized in  the `PodTemplate.yaml` file \o/

PR #2881, opened 2022-09-14
feat(artifact-caching-proxy): add AWS provider

lemeurherve:
Include the addition of nginx-ingress to cik8s

Testing with default gp2 storage class for now

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752


dduportal (approved):
Nice job! LGTM (under the assumption that you'll check the helmfile diff before merging + you might need to install ACME CRDS manually in cik8S first)

PR #2934, opened 2022-09-22
chore: delete jenkins-infra charts docker image manifests

lemeurherve:
This update is done in our corresponding helm charts, no need to duplicate here


dduportal (approved):
https://github.com/jenkins-infra/helm-charts/pull/224 these elements in the helm charts

LGTM

PR #2939, opened 2022-09-23
chore(artifact-caching-proxy): increase resources

lemeurherve:
Node capacity on
- prodpublick8s: Standard_D8s_v3, 8 vCPU 32Go mem
- doks-public: s-4vcpu-8gb, 4 vCPU 8Go mem
- eks-public: t4g.xlarge, 4 vCPU 16Go mem

### DIFF
--- a/config/artifact-caching-proxy__common.yaml
+++ b/config/artifact-caching-proxy__common.yaml
@@ -1,11 +1,11 @@
 # Common values for all nginx-proxies
 resources:
   limits:
-    cpu: 2
-    memory: 1024Mi
+    cpu: 3
+    memory: 6146Mi
   requests:
-    cpu: 2
-    memory: 1024Mi
+    cpu: 3
+    memory: 6146Mi
 
 persistence:
   enabled: true


timja (commented):
um does it really need that much?

where did these numbers come from?

is it just spare capacity on the node not being used for anything else?

config/artifact-caching-proxy__common.yaml
```diff
@@ -1,11 +1,11 @@
 # Common values for all nginx-proxies
 resources:
   limits:
-    cpu: 2
-    memory: 1024Mi
+    cpu: 3
+    memory: 6146Mi
```
dduportal, line 5: Will it work on the 3 clusters? (e.g. do you mind adding links to the 3 node size definition in the PR body)

lemeurherve: I've added the node capacity on the 3 clusters.

dduportal:
> um does it really need that much?
> 
> where did these numbers come from?
> 
> is it just spare capacity on the node not being used for anything else?

For the vCPU: it's nginx so 1 vCPU for the (master process + cache manager), and then 1vCPU per worker process.
The choice of 3 vCPU is to ensure that there are always 1 vCPU available for worker (when there is a reload, nginx duplicates the worker process with the new conf).

For the memory, usually 1 Gb per CPU is ok on a standard reverse proxy or webserver usage. But with the cache manager, better to get more memory to have some room so 2Gb per vCPU.

timja:
> For the memory, usually 1 Gb per CPU is ok on a standard reverse proxy or webserver usage. But with the cache manager, better to get more memory to have some room so 2Gb per vCPU.

has it been tested as needing that much?

dduportal:
> > For the memory, usually 1 Gb per CPU is ok on a standard reverse proxy or webserver usage. But with the cache manager, better to get more memory to have some room so 2Gb per vCPU.
> 
> has it been tested as needing that much?

- For the >1 vCPU, yes (we need at least 2).
- For the 3vCPUs, no, we can start at 2. But the 2 -> 3 difference does not matter as 2 of the 3 installations (EKS-public and DOKS-public) only have 1 worker node with only nginx within, and the 3rd cluster (AKS - prodpublick8s) had to scale up to fit even a 2CPU/2Gb service)
- For the 2 Gb per vCPU: yes on the initial pod on AKS.

dduportal (approved):

PR #2998, opened 2022-10-05
fix(wiki): remove image.tag from its values, tracked in helm chart

lemeurherve:


lemeurherve:
Note: we have the same kind of problems in some other values, those for our helm charts at first sight. Complementary PR incoming.

PR #2999, opened 2022-10-05
fix: don't pin image.tag here when they're already tracked in our helm charts

lemeurherve:
Follow-up of https://github.com/jenkins-infra/kubernetes-management/pull/2998

Put as draft as some remaining image.tag might need some reflections before removing them.


halkeye:
cds should :hocho: 
boo custom distribution service.

dduportal (dismissed):

dduportal (approved):

PR #3024, opened 2022-10-13
fix(eks-public / datadog):  add datadog release with amd64

smerle33:
after : https://github.com/jenkins-infra/aws/pull/250
datadog install with the amd64 image

PR #3025, opened 2022-10-13
fix(eks-public): restore nginx-ingress default backend

lemeurherve:
Now that we've switched from an incompatible ARM CPU architecture to a compatible amd64 one in https://github.com/jenkins-infra/aws/pull/250

Related restoration, same effect: #3024


smerle33 (approved):
🚀

PR #3027, opened 2022-10-13
fix(eks-public): correct instance type for the autoscaler node selector

lemeurherve:
As the instance type has changed from 't4g.xlarge' to 't3a.xlarge' since https://github.com/jenkins-infra/aws/pull/250, the autoscaler pod stayed in 'pending' with the following warning:
> Warning  FailedScheduling  14s (x2 over 104s)  default-scheduler  0/2 nodes are available: 2 node(s) didn’t match Pod’s node affinity/selector.


smerle33 (approved):
🚀

PR #3035, opened 2022-10-13
fix(eks-public) correct remaining configs

dduportal:


config/ext_aws-node-termination-handler_eks-public.yaml
```diff
@@ -3,4 +3,4 @@ awsRegion: us-east-2
 
 nodeSelector:
   node.kubernetes.io/lifecycle: normal
-  node.kubernetes.io/instance-type: t4g.xlarge
+  node.kubernetes.io/instance-type: t3a.xlarge
```
lemeurherve, line 6: 😅

PR #3036, opened 2022-10-14
Update ircbot to latest release

slide:
Updates ircbot to latest tag


lemeurherve (changes requested):
As @halkeye has added an updatecli manifest to track the IRCBot docker image version in https://github.com/jenkins-infra/helm-charts/pull/250 we can now use the new corresponding helm chart version to update it, an updatecli automatic pull request will take care of it.

Thanks anyway @slide !

PR #3065, opened 2022-10-19
chore(updatecli) fix manifests with dockerdigest in 0.36.x

dduportal:
Same thing as https://github.com/jenkins-infra/jenkins-infra/pull/2439.

The goal of this PR is to avoid duplication of the `sha256:` prefix due to updatecli v0.36.0


dduportal:
Merging as it it a fix to close #3066

PR #3197, opened 2022-11-10
fix(infra.ci): add command for new node agent

smerle33:
add a command to the node.

### DIFF
--- a/config/ext_jenkins-infra.yaml
+++ b/config/ext_jenkins-infra.yaml
@@ -221,6 +221,7 @@ controller:
                     containers:
                       - name: "jnlp"
                         image: "jenkinsciinfra/builder@sha256:230919e6a856373c43ed01c4f967cab6274599f6457cd1ea49850ea759e67aa4"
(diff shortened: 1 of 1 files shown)


config/ext_jenkins-infra.yaml
```diff
@@ -233,7 +234,7 @@ controller:
                             value: "~/.asdf/shims:~/.asdf/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
                         - envVar:
                             key: "ARTIFACT_CACHING_PROXY_PROVIDER"
-                            value: "do"
+                            value: "azure"
```
lemeurherve, line 237: Why this change?

dduportal: yep: the containers issued for infra.ci are all in the same cluster (temp-privatek8s)  e.g. inside Azure.

The initial value here comes from a copy and paste from ci.jenkins.io.

dduportal (approved):

PR #3194, opened 2022-11-10
feat: add `privatek8s` AKS cluster

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2844

<details><summary>previously:</summary>

Diff between ext_jenkins-infra.yaml & ext_jenkins-infra_privatek8s.yaml temporary introduced until the switch from temp_privatek8s and privatek8s cluster for infra.ci.jenkins.io:

```diff
--- ext_jenkins-infra.yaml
+++ ext_jenkins-infra_privatek8s.yaml
@@ -1,3 +1,4 @@
+## TODO: rename this file to ext_jenkins-infra.yaml and get rid of the old one when the switch from temp-privatek8s to privatek8s is done
 serviceAccount:
   create: true
   name: jenkins-controller
@@ -40,10 +41,6 @@
     readinessProbe:
       initialDelaySeconds: 120
   testEnabled: false
-  podSecurityContextOverride:
-    runAsNonRoot: true
-    runAsUser: 1000
-    supplementalGroups: [1000]
   overwritePlugins: true
   serviceType: "ClusterIP"
   javaOpts: >-
@@ -88,10 +85,10 @@
                     username: "MY-USERNAME"
                 - file:
                     fileName: "kubeconfig"
-                    id: "kubeconfig-controller-temp-privatek8s"
-                    description: "Kubeconfig file for temp-privatek8s"
+                    id: "kubeconfig-controller-privatek8s"
+                    description: "Kubeconfig file for privatek8s"
                     scope: SYSTEM
-                    secretBytes: "${base64:${KUBECONFIG_TEMP_PRIVATEK8S}}"
+                    secretBytes: "${base64:${KUBECONFIG_PRIVATEK8S}}"
                 - basicSSHUserPrivateKey:
                     scope: SYSTEM
                     id: "ec2-agents-privkey"
@@ -120,7 +117,7 @@
                 namespace: "jenkins-agents"
                 podRetention: "Never"
                 serverUrl: ${AGENTS_KUBERNETES_CLUSTER_URL}
-                credentialsId: "kubeconfig-controller-temp-privatek8s"
+                credentialsId: "kubeconfig-controller-privatek8s"
                 serverCertificate: |
                   ${AGENTS_KUBERNETES_CLUSTER_CERT}
                 podLabels:

```

</details>

**About the removal of `podSecurityContextOverride` block:**
[...]

PR #3233, opened 2022-11-18
chore: factorize nginx ingresses values

lemeurherve:
This pull request factorizes common values for the private and public nginx-ingress services running on all clusters.

Extracted from #3194

### DIFF
--- a/clusters/prodpublick8s.yaml
+++ b/clusters/prodpublick8s.yaml
@@ -46,6 +46,7 @@ releases:
     chart: ingress-nginx/ingress-nginx
     version: 4.4.0
     values:
+      - "../config/ext_private-nginx-ingress__common.yaml"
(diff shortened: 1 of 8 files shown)


dduportal (changes requested):
Blocking because the `helmfile diff` show thing we do not want.

LGTM on the principle, but need a few minor fixes (I'll add them a bit later today but use the diff Luke if you want to work on it earlier)

dduportal:
For temp-privatek8s:

```diff
-   whitelist-source-range: "140.82.112.0/20,192.30.252.0/22"`
```

dduportal:
For prodpublick8s:

```diff
-             - --ingress-class=private-ingress
+             - --ingress-class=nginx
```

dduportal:
@lemeurherve you can proceed to merge when you want: the only failing check is updatecli diff.
The helmfile diff is ok for me now.

PR #3326, opened 2022-12-12
fix(weekly.ci.jenkins.io) fix rbac: `anonymous` is a user, not a group

dduportal:
Related to #3323

Issue #3323, opened 2022-12-10
Anonymous is a user


daniel-beck:
Saw a form validation error browsing weekly.ci, caused by https://github.com/jenkins-infra/kubernetes-management/blob/c61cf1acf9b9c79439cc8d18df7b8a61856af591/config/ext_jenkins-weekly.yaml#L109

Anonymous is a user, not a group. (The user usually gets the group membership anyway so it doesn't make much of a difference in practice 🤷)

dduportal:
https://github.com/jenkins-infra/kubernetes-management/pull/3326 should fix this. Thanks @daniel-beck !

daniel-beck:
@dduportal Not fully: https://github.com/jenkins-infra/kubernetes-management/pull/3326/files#r1049608017

PR #3349, opened 2022-12-15
fix(weekly.jenkins.io): `anonymous` is a user not a group

lemeurherve:
Follow-up of #3326

Ref: https://github.com/jenkins-infra/kubernetes-management/issues/3323#issuecomment-1353022203


dduportal (approved):
Thanks @daniel-beck @lemeurherve !

PR #3487, opened 2023-01-24
feat: add `publick8s` cluster

lemeurherve:

PR #3514, opened 2023-01-26
feat(publick8s): add `artifact-caching-proxy`

lemeurherve:
Extract from #3487 


dduportal (changes requested):
Temporarly blocking: the principal branch of the job is failing with the private ingress not being installed properly.

Caught when reviewing the helmfile diff here: the private ingress is marked as "never installed before" which is unexpected

lemeurherve:
Yes I'm working on it.

dduportal (approved):

PR #3517, opened 2023-01-26
fix(eks-public/doks-public): upgrade cert-manager version to v1.11.0

smerle33:
updatecli was not tracking cert-manager versions for eks-public and doks-public.


lemeurherve (approved):
Great catch!

PR #3524, opened 2023-01-27
cleanup(prodpublick8s): stop tracking `grafana` & `prometheus`

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2844

Note: the following DNS records will have to be deleted from Azure:
- [ ] grafana.jenkins.io
- [ ] grafana.publick8s.jenkins.io

(edit by dduportal) also, the namespaces will have to be deleted (in pair) after this PR is applied.
I've updated the title to "stop tracking" to avoid confusion with "deletion" as merging this won't delete anything.


dduportal:
Setting as draft until we decide of a migration planning

dduportal (approved):

dduportal:
Approving as discussed with @olblak during FOSDEM 2023. Since we have full datadog metrics/logs collection, and since most of this stack is not used, better cleaning it up to avoid paying for nothing.

We can absolutley install a new on in the future if needed.

PR #3528, opened 2023-01-27
feat: deploy `falco` on every cluster

lemeurherve:
Increased security.


dduportal:
Setting as draft until we decide of a migration planning

lemeurherve:
As discussed together, this PR isn't related to the cluster migration, but as it can prevent pods creation it needs a cautious review so we'll keep it as draft for now.

dduportal:
This PR is full of conflicts, not planned and too old.

As much as we would want to increase security, we'll close it until we have an issue and a plan.

PR #3536, opened 2023-01-30
feat(artifact-caching-proxy): set two replicas on every provider

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2752


dduportal (approved):
LGTM 👍

PR #3537, opened 2023-01-31
fix(artifact-caching-proxy): distinct health check path

lemeurherve:
As the `/healthz` path is the default health check path of the nginx ingress, this PR is specifying a distinct path to ensure the artifact-caching-proxy health is tested and not the nginx ingress.

Note: [the default chart value](https://github.com/jenkins-infra/helm-charts/blob/4173bbb9d8238ab89be27085f09234459a1298a2/charts/artifact-caching-proxy/values.yaml#L43) could also be changed.

### DIFF
--- a/config/artifact-caching-proxy__common.yaml
+++ b/config/artifact-caching-proxy__common.yaml
@@ -14,6 +14,7 @@ persistence:
 ingress:
   enabled: true
   className: public-nginx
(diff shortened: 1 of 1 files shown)


dduportal (approved):
LGTM.

That would improve the monitoring of the ACP service \o/

PR #3538, opened 2023-01-31
fix(eks-public) correct ACP and public-ingress configuration to follow cert-manager recommendations

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3305, this PR persists the manual changes that were done to allow Cert Manager to validate the HTTP-01 Let's Encrypt challenge in eks-public, for the ACP service.

- Setting the ingress class `public-nginx`as the default (to pick and implement any ingress rule without a `spec.ingressClass` attribute set to the value `public-nginx`
- Enabling the class by name mode to ensure that all cases of ingress rule specification are covered
- Set up ACP ingress rule to reuse its Ingress rule "in place" (instead of creating/deleting another ingress rule each time a challenge happens)

References:
- https://github.com/cert-manager/cert-manager/issues/4821
- https://cert-manager.io/docs/installation/upgrading/ingress-class-compatibility/

### DIFF
--- a/config/artifact-caching-proxy_aws.yaml
+++ b/config/artifact-caching-proxy_aws.yaml
@@ -1,4 +1,7 @@
 ingress:
+  enabled: true
+  annotations:
+    acme.cert-manager.io/http01-edit-in-place: "true"
   hosts:
     - host: repo.aws.jenkins.io
       paths:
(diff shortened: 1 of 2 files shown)


config/artifact-caching-proxy_aws.yaml
```diff
@@ -1,4 +1,7 @@
 ingress:
+  enabled: true
```
lemeurherve, line 2:
```suggestion
```
Already in the __common values

lemeurherve (approved):
I would put the URL in the PR body instead of in its title, LGTM otherwise

PR #3648, opened 2023-03-02
feat(updatecli): add the jenkins-infra (puppet) entry to updatecli

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3419 
this PR add the repository jenkins-infra with the jenkins file `Jenkinsfile_updatecli` to the infra.ci.jenkins.io in order to replace the GHA.
The `Jenkinsfile_updatecli` is provided by the PR : https://github.com/jenkins-infra/jenkins-infra/pull/2673

PR #3655, opened 2023-03-03
fix(infra.ci) ensure that the updatecli job for puppet (jenkins-infra) watches the correct branches (production, etc.)

dduportal:
Fixup of #3648


smerle33 (approved):
🚀

PR #3660, opened 2023-03-06
feat(cik8s,doks) increase the max. pod quotas for jenkins-agents namespaces

dduportal:
- Reference for cik8s: https://github.com/jenkins-infra/jenkins-infra/blob/b2ed0ee76d2d362a51353fa511a86b6792c10cd8/hieradata/clients/azure.ci.jenkins.io.yaml#L148
- Reference for doks: https://github.com/jenkins-infra/jenkins-infra/blob/b2ed0ee76d2d362a51353fa511a86b6792c10cd8/hieradata/clients/azure.ci.jenkins.io.yaml#L247

(edit) this PR follows up https://github.com/jenkins-infra/jenkins-infra/pull/2655 which changed the quotas on the Jenkins Kuberenetes plugin side. I forgot to propagate the change manually on Kubernetes side.

PS: this change will be automated by an updatecli manifest soon (ref. jenkins-infra/helpdesk#3431)

PR #3661, opened 2023-03-07
feat(doks) increase the max. pod quotas for jenkins-agents namespaces

lemeurherve:
As the droplets limit on DigitalOcean account has been increased to 100, we can now increase this cluster capacity to be the same as the AWS one.

Follow-up of https://github.com/jenkins-infra/jenkins-infra/pull/2679 & #3660


smerle33 (approved):
🚀

PR #3663, opened 2023-03-08
chore(pods quotas): move quotas as values

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3431 
to allow an easier updatecli way of tracking the quotas, this move the quotas.pods to a values file


smerle33:
as per the error : 
```
[2023-03-08T07:58:04.960Z] in clusters/doks.yaml: error during doks.yaml.part.0 parsing: template: stringTemplate:48:26: executing "stringTemplate" at <.quotas.pods.doks>: can't evaluate field quotas in type *state.EnvironmentTemplateData

script returned exit code 1
```

I probably need to check that the values file is "loaded" first

dduportal (changes requested):
You are mixing helm (the template engine with its `{{ .Values ... }}` syntax, with helmfile (the deployment tool which uses helm and sops).


If you look at the other releases in the cluster manifest files, you'll see that there are using a `values` directive to point to the relative path of the values file. (also https://helmfile.readthedocs.io/en/latest/ can help if not clear).

Helmfile provides tempalting with `{{ .Values }}` as well but we do not use it: WDYT about keeping the same pattern as other services and add 2 `config/***.yaml` files instead (1 per cluster which need a value)?

lemeurherve (dismissed):
LGTM

Nit: maybe rename the values files to `config/jenkins-kubernetes-agents_<cluster-name>.yaml` so we immediately know they are related to the `jenkins-kubernetes-agents` helm chart, WDYT?

clusters/doks.yaml
```diff
@@ -43,6 +43,5 @@ releases:
     version: 0.4.0
     secrets:
       - "../secrets/config/jenkins-kubernetes-agents/secrets.yaml"
-    set:
-      - name: "quotas.pods"
-        value: "150"
+    values:
+      - "../config/jenkins-kubernetes-agents_cik8s.yaml"
```
dduportal, line 47:
```suggestion
      - "../config/jenkins-kubernetes-agents_doks.yaml"
```

😅

dduportal (changes requested):
LGTM, except one (blocking) typo ! Nice job 👏

smerle33:
> LGTM, except one (blocking) typo ! Nice job 👏

oops

dduportal (approved):

lemeurherve (approved):

PR #3683, opened 2023-03-09
chore: move `release.ci.jenkins.io` from `prodpublick8s` cluster to `privatek8s` cluster

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2844


config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,689 @@
...
+      #               value: "amd64"
+      #             - name: "os"
+      #               value: "windows"
+      #             - name: "jenkins"
+      #               value: "release.ci.jenkins.io"
+      #             tenancy: Default
+      #             tmpDir: "C:\\\\Temp"                          # Mandatory because EC2 plugins assumes SSH = Unix and searches for /tmp
+      #             type: T3Medium
+      #             useEphemeralDevices: true
+      jobs-settings: |
```
dduportal, line 393: Can we start with an "empty" state at first, to really avoid any "unwanted" operation ? (e.g. commenting out the jobs as well)

lemeurherve:
Testing this PR in its current state from my machine first:
```shell
$ kubectl config use-context privatek8s-emerging-ram
Switched to context "privatek8s-emerging-ram".

$ helmfile apply -f clusters/privatek8s.yaml -l name=jenkins-release     
```

config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,692 @@
...
+      initialDelaySeconds: 120
+    livenessProbe:
+      initialDelaySeconds: 120
+    readinessProbe:
+      initialDelaySeconds: 120
+  testEnabled: false
+  podSecurityContextOverride:
+    runAsNonRoot: true
+    runAsUser: 1000
+    supplementalGroups: [1000]
```
lemeurherve, line 54:
```suggestion
```
I wonder if we should or not remove this part, not present in the infra values.

WDYT?

config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,692 @@
...
+  size: 50Gi
+agent:
+  componentName: "agent"
+networkPolicy:
+  # As of today, 2020-06-19, network policy is not supported for windows node
+  # https://docs.microsoft.com/en-us/azure/aks/windows-node-limitations#are-all-features-supported-with-windows-nodes
+  # This is a blocker for the release environment as in the current state,
+  # windows containers are not allowed to connect on main Jenkins.
+  # I disable it for now
+  # enabled: true
```
lemeurherve, line 23: To be checked if it's still the case.

config/ext_jenkins-release_privatek8s.yaml
lemeurherve: Removed as it prevented the init container to write on the PV

config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,692 @@
...
+      memory: "4Gi"
+  probes:
+    startupProbe:
+      initialDelaySeconds: 120
+    livenessProbe:
+      initialDelaySeconds: 120
+    readinessProbe:
+      initialDelaySeconds: 120
+  testEnabled: false
+  numExecutors: 0
```
lemeurherve, line 52: [async] To be reported on infra & weekly?

config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,692 @@
...
+            - "GCLogs"
+            - "Agents"
+            - "RootCAs"
+            - "SlaveLogs"
+            - "HeapUsageHistogram"
+          nagDisabled: true
+      jenkins-url: |
+        unclassified:
+          location:
+            url: "https://release.ci.jenkins.io"
```
lemeurherve, line 640: Added to please the warning shown in Jenkins UI

[async] To be reported on infra & weekly?

config/ext_jenkins-release_privatek8s.yaml
dduportal: This line is not needed: 

- Current configuration sets `controller.JCasc.defaultConfig` to `false` so the `controller.numExecutors` is never used (https://github.com/jenkinsci/helm-charts/blob/12a3c0992b9be68490e8944211506a22c5814b5f/charts/jenkins/templates/jcasc-config.yaml#L25-L45 and https://github.com/jenkinsci/helm-charts/blob/aff05875d84a7d79f010abc8756a11940804131b/charts/jenkins/templates/config.yaml#L60-L62 , the if is false so the template for the jcasc "defaults" which [consumes `numExecutors`](https://github.com/jenkinsci/helm-charts/blob/d0aa61f5921fb937e6f0a75ee3c1cf1f86cd2b18/charts/jenkins/templates/_helpers.tpl#L104-L121) is never called)

dduportal: besides, if we keep the config in the JCasc configuration, it will always override

config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,692 @@
+serviceAccount:
+  create: true
+  name: jenkins-controller
+# serviceAccountAgent:
+#   create: true
+#   name: jenkins-agent
```
lemeurherve, line 6:
```suggestion
```
Old values

config/ext_jenkins-release_privatek8s.yaml
```diff
@@ -0,0 +1,692 @@
...
+      #               value: "release.ci.jenkins.io"
+      #             tenancy: Default
+      #             tmpDir: "C:\\\\Temp"                          # Mandatory because EC2 plugins assumes SSH = Unix and searches for /tmp
+      #             type: T3Medium
+      #             useEphemeralDevices: true
+      jobs-settings: |
+        jobs:
+          - script: >
+              folder('components') {
+                displayName('Components (on privatek8s)')
```
lemeurherve, line 396: Temporary marker to be sure of the instance, to be removed:
```suggestion
                displayName('Components')
```

config/ext_jenkins-release_privatek8s.yaml
lemeurherve: Yup, it didn't do anything on a new apply, the (empty) release controller had 2 executors, reverted this move.

lemeurherve: > Current configuration sets controller.JCasc.defaultConfig to false so the controller.numExecutors is never used ([jenkinsci/helm-charts@12a3c09/charts/jenkins/templates/jcasc-config.yaml#L25-L45](https://github.com/jenkinsci/helm-charts/blob/12a3c0992b9be68490e8944211506a22c5814b5f/charts/jenkins/templates/jcasc-config.yaml#L25-L45) and [jenkinsci/helm-charts@aff0587/charts/jenkins/templates/config.yaml#L60-L62](https://github.com/jenkinsci/helm-charts/blob/aff05875d84a7d79f010abc8756a11940804131b/charts/jenkins/templates/config.yaml#L60-L62) , the if is false so the template for the jcasc "defaults" which [consumes numExecutors](https://github.com/jenkinsci/helm-charts/blob/d0aa61f5921fb937e6f0a75ee3c1cf1f86cd2b18/charts/jenkins/templates/_helpers.tpl#L104-L121) is never called)

I'll probably make a PR on the official chart to add a notice about that, not obvious when reading https://github.com/jenkinsci/helm-charts/blob/12a3c0992b9be68490e8944211506a22c5814b5f/charts/jenkins/VALUES_SUMMARY.md

dduportal (approved):

PR #3719, opened 2023-03-22
chore(cik8s,doks): rename jenkins-kubernetes-agents config files

lemeurherve:
Follow-up of #3718


dduportal (approved):
LGTM!

Helmfile diff reports no changes: 🚀

PR #3734, opened 2023-03-24
feat(privatek8s): track GitHub hooks IPs

lemeurherve:


config/ext_public-nginx-ingress_privatek8s.yaml
```diff
@@ -1,6 +1,6 @@
 controller:
   config:
-    ## Only allow GitHub's webhooks requests - https://github.blog/changelog/2019-04-09-webhooks-ip-changes/
+    # Only allow GitHub's webhooks requests - https://api.github.com/meta ("hooks", tracked by updatecli)
     whitelist-source-range: "140.82.112.0/20,192.30.252.0/22"
```
lemeurherve, line 4: Looking at the nginx chart doc, this annotation accepts CIDR: https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/configmap.md#whitelist-source-range

This doc also links to https://nginx.org/en/docs/http/ngx_http_access_module.html which shows IPv4 and IPv6 addresses together.

dduportal (approved):

PR #3760, opened 2023-03-28
chore(infra.ci/updatecli): add Azure version from packer image version in updatecli manifest

smerle33:
Still missing a condition that check the existence of the image within the azure gallery of packer-image with  
`az image list or show`


dduportal (approved):
LGTM 👍

PR #3816, opened 2023-04-07
feat(privatek8s) enable load balancer deny list for the public ingress, in full IPv4

dduportal:
While working on the migration https://github.com/jenkins-infra/helpdesk/issues/2844, @lemeurherve and I realized that the IP restriction of the public nginx ingress on the `privatek8s` cluster was only at the NBginx level.

It works as expected, but we can improve this with this PR, by ensuring that the IP denylist is also set at the Loadbalancer level.

The command 

```
curl --silent --verbose --insecure https://20.15.118.70 --output /dev/null 
```

should timeout once deployed, unless run from an authorized IP.

- The value `co,ntroller.service.loadBalancerSourceRanges` is documented in https://github.com/kubernetes/ingress-nginx/blob/26d83d1b2058cc9c23657b4e97b9e5a6d5fe9871/charts/ingress-nginx/README.md?plain=1#L292

- We already use it on the `prodpublick8s` cluster (for LDAP: https://github.com/jenkins-infra/helm-charts/blob/c1c01ce01a23d6c23861b3eebdfcf7b005376c08/charts/ldap/templates/service.yaml#L13): the AKS LoadBalancers are working properly with it

- This PR has been tested manually: the attribute `loadBalancerSourceRanges` was defined manually and the `curl` command timed out as expected. This manual change was rollbacked

- The updatecli manifest is failing in this PR but was tested locally without the `scmid/action`.
  - Note that only IPv4 is used, the IPv6 are filtered out: the network setup on `privatek8s` cluster does not allow for IPv6 at all.
  - I had to "cheat" by using a "file matchpattern" to define the YAMl list from the collection returned by updatecli. Ping @olblak if you have a better solution (my source is a string that can be split by `,`, my target is a YAML array)

### DIFF
--- a/config/public-nginx-ingress_privatek8s.yaml
+++ b/config/public-nginx-ingress_privatek8s.yaml
@@ -1,10 +1,11 @@
 controller:
   config:
     # Only allow GitHub's webhooks requests - https://api.github.com/meta ("hooks", tracked by updatecli)
-    whitelist-source-range: "140.82.112.0/20,143.55.64.0/20,185.199.108.0/22,192.30.252.0/22,2606:50c0::/32,2a0a:a440::/29"
+    whitelist-source-range: "140.82.112.0/20,143.55.64.0/20,185.199.108.0/22,192.30.252.0/22"
   service:
     annotations:
       service.beta.kubernetes.io/azure-load-balancer-internal: false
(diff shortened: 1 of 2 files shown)


lemeurherve (approved):
Nice tricks!

PR #3862, opened 2023-04-19
hotfix(prodpublick8s) migrate mirrorbits PV to native CSI, use a 1000 Gib size and fine-tune mountoptions

dduportal:
- Current used disk space is 486 Gb which is over the 80% threshold with a 500 Gib volume => setting it to `1000 Gib`
- Switch to CSI-premium tier to ensure performances
- Use CSI native options to eavoid any kind of "autoamtic" migration
- Fine tuned the mount options: only the owner should be able to write

### DIFF
--- a/config/mirrorbits.yaml
+++ b/config/mirrorbits.yaml
@@ -56,29 +56,33 @@ repository:
     spec:
       accessModes:
         - ReadWriteMany
-      storageClassName: azurefile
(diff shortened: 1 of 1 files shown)


dduportal:
This PR is a hotfix related to https://github.com/jenkins-infra/helpdesk/issues/3525, so no checks as it was applied successuflyy from a local machine

PR #3897, opened 2023-05-03
chore(cik8s,doks): switch to official Datadog Docker image without custom checks

lemeurherve:
As we recently encountered a lot of ["Weird response time" alerts](https://github.com/jenkins-infra/datadog/blob/bd70e30624fe58d4b1a75d045468febb0da3e0a7/datadog-monitors.tf#L93-L121) from `cik8s` cluster (probably due to a DNS problem), this PR switches to official Datadog Docker image for all clusters by default, and use [our own Datadog image with custom checks](https://github.com/jenkins-infra/docker-datadog/blob/main/conf.d/http_check.d/jenkins.yaml) on public facing clusters.

### DIFF
--- a/config/datadog_cik8s.yaml
+++ b/config/datadog_cik8s.yaml
@@ -11,9 +11,3 @@ agents:
       operator: "Equal"
       value: "true"
       effect: "NoSchedule"
-  image:
-    # official image without custom checks
-    repository: "datadog/agent"
-    tag: "7"
-    # we don't have to skip version test as we're not using the docker image digest
-    doNotCheckTag: false
--- a/config/datadog_cik8s.yaml
+++ b/config/datadog_cik8s.yaml
@@ -11,3 +11,9 @@ agents:
       operator: "Equal"
       value: "true"
       effect: "NoSchedule"
+  image:
+    # official image without custom checks
+    repository: "datadog/agent"
+    tag: "7"
+    # we don't have to skip version test as we're not using the docker image digest
+    doNotCheckTag: false
--- a/config/datadog_doks-public.yaml
+++ b/config/datadog_doks-public.yaml
@@ -4,7 +4,7 @@ clusterAgent:
   image:
     pullSecrets:
       - name: "dockerhub-credential"
-agent:
+agents:
   image:
     # repo & '@sha256' suffix here as we're retrieving a docker digest from updatecli
     repository: "jenkinsciinfra/datadog@sha256"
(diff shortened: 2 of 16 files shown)


config/datadog_cik8s.yaml
```diff
@@ -11,3 +11,9 @@ agents:
       operator: "Equal"
       value: "true"
       effect: "NoSchedule"
+  image:
+    # official image without custom checks
+    repository: "datadog/agent"
+    tag: "7"
+    # we don't have to skip version test as we're not using the docker image digest
+    doNotCheckTag: false
```
dduportal, line 19: Is there a reason to not pick the default values from the chart, aka. https://github.com/DataDog/helm-charts/blob/93c4fe5f2e58537a25626bee9656a21ef30415d8/charts/datadog/values.yaml#L1191-L1212 ?

dduportal (dismissed):
LGTM under the assumption that the override here is temporary and should be inverted (e.g. in long term, we should not change images unless we want the custom image)

lemeurherve:
I've inversed the logic: by default official Datadog image, custom image on public clusters only

config/datadog_doks-public.yaml
```diff
@@ -4,3 +4,10 @@ clusterAgent:
   image:
     pullSecrets:
       - name: "dockerhub-credential"
+agent:
```
dduportal, line 7:
```suggestion
agents:
```

config/datadog_eks-public.yaml
```diff
@@ -4,3 +4,10 @@ clusterAgent:
   image:
     pullSecrets:
       - name: "dockerhub-credential"
+agent:
```
dduportal, line 7:
```suggestion
agents:
```

config/datadog_prodpublick8s.yaml
```diff
@@ -8,3 +8,10 @@ datadog:
     hostCAPath: /etc/kubernetes/certs/kubeletserver.crt
     # Required as of Agent 7.35 because Kubelet certificates in AKS do not have a Subject Alternative Name (SAN) set.
     tlsVerify: false
+agent:
```
dduportal, line 11:
```suggestion
agents:
```

config/datadog_publick8s.yaml
```diff
@@ -8,3 +8,10 @@ datadog:
     hostCAPath: /etc/kubernetes/certs/kubeletserver.crt
     # Required as of Agent 7.35 because Kubelet certificates in AKS do not have a Subject Alternative Name (SAN) set.
     tlsVerify: false
+agent:
```
dduportal, line 11:
```suggestion
agents:
```

dduportal (changes requested):
Almost there! typo ;)

Issue #3901, opened 2023-05-05
The `jnlp-windows` agent definition of release.ci.jenkins.io references an inexistant Docker image


lemeurherve:
https://github.com/jenkins-infra/kubernetes-management/blob/1f2fa76a12e7bc1a0d5cb4cc48a4ef0298cf4232/config/jenkins_release.ci.jenkins.io.yaml#L243

This image doesn't exist:

```console
C:\> docker pull jenkinsciinfra/inbound-agent:windowsservercore-1809
Error response from daemon: manifest for jenkinsciinfra/inbound-agent:windowsservercore-1809 not found: manifest unknown: manifest unknown
```

PR #3914, opened 2023-05-09
cleanup(release.ci) change default (unused) windows inbound agent container image

dduportal:
Follow up of https://github.com/jenkins-infra/docker-inbound-agents/pull/77

The `jnlp-windows` base pod template is not used by the release jobs (they have their own yaml definition) but it's a nice feature to provide a base working image to allow extending from in the future => that's why this PR updates the image instead of removing the template (as you suggested to not remove it)


lemeurherve (approved):
Fixes https://github.com/jenkins-infra/kubernetes-management/issues/3901

PR #3916, opened 2023-05-10
cleanup(infra.ci): remove AWS arm64 vm replaced by azure ones

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3471
removing AWS arm64 vm

### DIFF
--- a/config/jenkins_infra.ci.jenkins.io.yaml
+++ b/config/jenkins_infra.ci.jenkins.io.yaml
@@ -72,20 +72,6 @@ controller:
                     password: "${ARTIFACT_CACHING_PROXY_PASSWORD}"
                     scope: GLOBAL
                     username: "MY-USERNAME"
-                - basicSSHUserPrivateKey:
-                    scope: SYSTEM
-                    id: "ec2-agents-privkey"
-                    username: "key"
-                    description: "SSH privkey used to connect to EC2 agents"
-                    privateKeySource:
-                      directEntry:
-                        privateKey: "${EC2_AGENTS_PRIVKEY}"
-                - aws:
-                    accessKey: "${EC2_AWS_ACCESS_KEY_ID}"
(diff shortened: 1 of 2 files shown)


config/jenkins_infra.ci.jenkins.io.yaml
```diff
@@ -226,49 +226,6 @@ controller:
                         resourceRequestMemory: "8G"
                     label: "container kubernetes node ruby webbuilder"
                     yamlMergeStrategy: "merge"
-            - amazonEC2:
-                cloudName: "aws-us-east-2"
-                credentialsId: "ec2-agents-credentials"
```
dduportal, line 231: Credential to remove from the top-level list of credential (same file)

config/jenkins_infra.ci.jenkins.io.yaml
```diff
@@ -226,49 +226,6 @@ controller:
...
-                instanceCapStr: "50"
-                noDelayProvisioning: true
-                region: "us-east-2"
-                sshKeysCredentialsId: "ec2-agents-privkey"
```
dduportal, line 235: Credential to remove from the top-level list of credential (same file)

dduportal (changes requested):
Nice work!

Could you also remove the 2 credentials (token used to access AWS API, and SSH key used to access the agents with SSH), unless you find another occurence.


No need to remove the secrets associated to these credentials though: better to wait for this change to be deployed

dduportal:
Nit: I would change `refactor`in the title by `cleanup` as it is not a refactor

smerle33:
> Nit: I would change `refactor`in the title by `cleanup` as it is not a refactor

I inspired my title with https://stackoverflow.com/questions/48075169/semantic-commit-type-when-remove-something

dduportal (approved):

PR #3933, opened 2023-05-13
Restore example jobs on weekly.ci

NotMyFault:
I checked something with design library this morning and noticed all jobs on weekly.ci are gone, and the header used to be plain text. Mark restored the previous header layout by flipping the security switch from plain text to safe html.

Therefore, I assume something (probably unintended) did happen to weekly.ci, I'm not aware of 👀 

A quick glance over the permission matrix on weekly.ci reveals I'm not permitted to modify jobs through the UI. Hence, I chose to come up with this JCasC approach, based on what I remember we had on weekly.ci.


dduportal:
Hello @NotMyFault good catch, thanks!

We migrated weekly.ci to another cluster friday (as per https://github.com/jenkins-infra/helpdesk/issues/3351#issuecomment-1546026790) and it seems some of the configuration was not in the Iac.

Thanks for this PR, let's start with this one!


Btw, you can still access the former instance by updating your `/etc/hosts` if needed!

```
curl -k -H 'Host: weekly.ci.jenkins.io' https://52.167.253.43
```


(cc @lemeurherve for info : we should avoid deleting the former instance until the new one is "ok" as confirmed by @NotMyFault )

dduportal (approved):

PR #3951, opened 2023-05-21
Respect users' os theme on weekly.ci

NotMyFault:
If you're a darkmode user and not logged into weekly.ci, your os theme set is now respected properly.

### DIFF
--- a/config/jenkins_weekly.ci.jenkins.io.yaml
+++ b/config/jenkins_weekly.ci.jenkins.io.yaml
@@ -121,7 +121,7 @@ controller:
           defaultFolderConfiguration:
             # Keep healthMetrics an empty list to ensure weather is disabled
             healthMetrics: []
-          theme-manager:
+          themeManager:
             disableUserThemes: false
             theme: "darkSystem"
         jenkins:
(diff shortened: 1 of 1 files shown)


config/jenkins_weekly.ci.jenkins.io.yaml
```diff
@@ -121,6 +121,9 @@ controller:
           defaultFolderConfiguration:
             # Keep healthMetrics an empty list to ensure weather is disabled
             healthMetrics: []
+          theme-manager:
```
timja, line 124: https://github.com/jenkinsci/theme-manager-plugin/blob/280bd2f83dc695f11c4b754f2dd9888cfdb9dacc/src/main/java/io/jenkins/plugins/thememanager/ThemeManagerPageDecorator.java#L21
```suggestion
          themeManager:
```

NotMyFault: Oh I could swear I copied that

timja (approved):

dduportal (approved):

PR #3977, opened 2023-05-24
chore(weekly.j.ci): remove emoticon from title that crash updatecli yaml parser

smerle33:
🚀  as \U0001F680


dduportal (approved):
Let's merge right now.

For info, should be solved in updatecli by https://github.com/updatecli/updatecli/pull/1333

PR #3945, opened 2023-05-17
chore(wiki): migrate from `prodpublick8s` to `publick8s`

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3351

PR #3983, opened 2023-05-24
feat: migrate keycloak management from prodpublick8s to publick8s

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3351, this PR is the first step to the migration of Keycloak to the new cluster `publick8s`.

It introduces the following changes:

- Stop managing keycloak in `prodpublick8s`
- Add the keycloak "admin" backend only in `publick8s` 

Please note that the resulting instance will fail at first:

- The database configuration is not updated yet (still points at the old version)
- No "public" access (the ingress `beta.accounts.jenkins.io` created by the jenkins-infra/keycloak-public release was removed 2 years ago)
- LDAP should allow requests from the new keycloak but to be verified


clusters/publick8s.yaml
```diff
@@ -111,3 +114,13 @@ releases:
...
       - "../config/wiki.yaml"
+  - name: keycloak # Admin: https://admin.accounts.jenkins.io/auth/admin
+    namespace: keycloak
+    chart: codecentric/keycloak
+    version: 18.4.3
+    timeout: 600
+    values:
+      - "../config/keycloak.yaml"
+    secrets:
+      - "../secrets/config/keycloak/db.yaml"
```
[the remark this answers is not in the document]
lemeurherve:
```suggestion
      - "../secrets/config/keycloak/public-db.yaml"
```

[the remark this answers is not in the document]
dduportal: good point, let me do it

clusters/prodpublick8s.yaml
```diff
@@ -146,23 +146,6 @@ releases:
...
-    secrets:
-      - "../secrets/config/keycloak/db.yaml"
-      - "../secrets/config/keycloak/http.yaml"
-  - name: keycloak-public # User: https://admin.accounts.jenkins.io/auth/realms/jenkins/account
-    namespace: keycloak
-    chart: jenkins-infra/keycloak
-    version: 0.1.1
-    timeout: 150
-    values:
-      - "../config/keycloak-public.yaml"
```
lemeurherve, line 165: Maybe remove keycloak-public in a separate PR?

clusters/prodpublick8s.yaml
lemeurherve, line 1: The repositories section should be updated here too by removing L7 to L9

clusters/publick8s.yaml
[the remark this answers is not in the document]
dduportal: Done (and added the new secret in https://github.com/jenkins-infra/charts-secrets/commit/9adca8ec9b4c925ca67c51d1377860dad5a96005)

clusters/prodpublick8s.yaml
dduportal: I opeted for the same pattern as in https://github.com/jenkins-infra/kubernetes-management/pull/3945. We can split in 2 not a problem

lemeurherve: The deletion of keycloak-public is distinct from the migration of keycloak (admin), not exactly the same purpose.

As you like 🙂

lemeurherve: I would keep the deletion of the keycloak (admin) release from clusters/prodpublick8s.yaml in this PR in any case.

dduportal: Oh I see, another incoming update.

A 2nd PR only for removing the "public" would make sense for auditability, while "single" PR make sense for "migration" so both are OK for me.

clusters/prodpublick8s.yaml
dduportal: Done

clusters/prodpublick8s.yaml
dduportal: https://github.com/jenkins-infra/kubernetes-management/pull/3984 here you are

lemeurherve (approved):
🚀

PR #3991, opened 2023-05-25
Enable context aware headers on weekly.ci

NotMyFault:
Displays different headers based on the page you are at.

### DIFF
--- a/config/jenkins_weekly.ci.jenkins.io.yaml
+++ b/config/jenkins_weekly.ci.jenkins.io.yaml
@@ -137,7 +137,7 @@ controller:
             logo: "default"
             logoText: "Jenkins"
             title: "with the <a href=\"https://plugins.jenkins.io/customizable-header/\">customizable-header</a>\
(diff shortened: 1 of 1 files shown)


config/jenkins_weekly.ci.jenkins.io.yaml
```diff
@@ -126,13 +126,18 @@ controller:
...
+                showJobWeather: true
             headerColor:
               backgroundColor: "#3B3B3B"
               color: "white"
               hoverColor: "grey"
             logo: "default"
             logoText: "Jenkins"
-            title: "with the customizable-header plugin"
+            title: "with the <a href=\"https://plugins.jenkins.io/customizable-header/\">customizable-header</a>\
+              \ plugin 🚀"
```
dduportal, line 140:
```suggestion
              \ plugin"
```

We have to avoid emoji and unicode for now (it breaks the updatecli updates)

dduportal (approved):

PR #4024, opened 2023-06-01
fix(weekly.ci.jenkins.io): replace `pullPolicy` field name by `imagePullPolicy`

lemeurherve:
There is no `pullPolicy` value property in https://github.com/jenkinsci/helm-charts/blob/main/charts/jenkins/values.yaml


dduportal (approved):
Good catch!

PR #4028, opened 2023-06-02
chore(reports): migrate from `prodpublick8s` to `publick8s`

lemeurherve:


config/reports.yaml
```diff
@@ -16,7 +16,7 @@ ingress:
       hosts:
         - reports.jenkins.io
         - reports.jenkins-ci.org
```
lemeurherve, line 18: Note: there is no associated record in the `jenkins-ci.org` DNS zone.

To be removed in a later PR?

clusters/plugin-health-scoring.yaml
```diff
@@ -0,0 +1,10 @@
+  - name: plugin-health-scoring
+    namespace: plugin-health-scoring
+    chart: jenkins-infra/plugin-health-scoring
+    version: 2.1.5
+    needs:
+      - public-nginx-ingress/public-nginx-ingress # Required to expose the service
+    values:
+      - "../config/plugin-health-scoring.yaml"
+    secrets:
+      - "../secrets/config/plugin-health-scoring/secrets.yaml"
```
dduportal, line 10: Looks like an error (`git add .` is bad :troll: )

dduportal (changes requested):
Accidental file commited, otherwise LGTM

clusters/plugin-health-scoring.yaml
lemeurherve: That was a `git add clusters`, completely unrelated! 😛 

Good catch, that's from a POC trying to allow external contributor to be in charge of dependency PR review for specific services

dduportal (approved):

PR #4038, opened 2023-06-05
chore(publick8s): move javadoc to arm node pool

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3584
moving javadoc to this new pool


dduportal (approved):
Looks good, but needs to check the deployement closely: @smerle33 I let you merge when you are ready to monitor the deployment and the pod events

PR #4042, opened 2023-06-06
chore(ldap): deploy to `publick8s`

lemeurherve:
This PR adds the LDAP service to `publick8s` cluster, with a new secret configuration using the new storage account with restricted access created in https://github.com/jenkins-infra/azure/pull/392 instead of the existing one which will be removed at the end of the migration described in https://github.com/jenkins-infra/helpdesk/issues/3351#issuecomment-1576741071


dduportal (approved):
👍

PR #4044, opened 2023-06-06
chore(ldap): deploy to `publick8s` and set its load balancer to use a public IP defined as code

lemeurherve:
Same as #4042 but without modifying the existing LDAP release on `prodpublick8s`.

> This PR adds the LDAP service to `publick8s` cluster, with a new secret configuration using the new storage account with restricted access created in https://github.com/jenkins-infra/azure/pull/392 instead of the existing one which will be removed at the end of the migration described in https://github.com/jenkins-infra/helpdesk/issues/3351#issuecomment-1576741071

### DIFF
--- a/config/ldap_publick8s.yaml
+++ b/config/ldap_publick8s.yaml
@@ -41,3 +41,6 @@ resources:
   requests:
     cpu: 500m
     memory: 1536Mi
+
+nodeSelector:
+  agentpool: x86medium
--- a/config/ldap_publick8s.yaml
+++ b/config/ldap_publick8s.yaml
@@ -1,5 +1,7 @@
 service:
   type: LoadBalancer
+  # Public IPv4 defined as code in https://github.com/jenkins-infra/azure/blob/main/ldap.jenkins.io.tf instead of requesting a random new public IP, useful for DNS setup and changes
+  IP: 104.209.245.168
   whitelisted_sources:
     - '10.0.32.0/19'  # Kubernetes publick8s internal IPs
     - '73.71.177.172/32'  # 106 accept inbound LDAPS request from spambot
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -177,3 +177,13 @@ releases:
       - "../config/accountapp.yaml"
     secrets:
       - "../secrets/config/accountapp/secrets.yaml"
+  - name: ldap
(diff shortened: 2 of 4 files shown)


lemeurherve:
helmfile diff OK

config/ldap_publick8s.yaml
```diff
@@ -0,0 +1,46 @@
+service:
+  type: LoadBalancer
```
dduportal, line 2: A thought about this LoadBalancer: the helm chart allows specifying a predefined IP through the `service.IP` value (ref. https://github.com/jenkins-infra/helm-charts/blob/4dc496fe3f2d17f399272ab6fdd6903083031089/charts/ldap/values.yaml#L58).

- You could define an Azure Public IP resource in jenkins-infra/azure which value could be specified here. Then, you could use jenkins-infra/azure (instead of azure-net) to have a DNS record which value is set from the IP.
  - This case makes the IP deterministic but MUST be done before the initial creation of the LB, because it cannot be changed later.
- Otherwise you'll have to wait for the release to be installed and the LB allocated, to retrieve the IP of the LB adn set the DNS `A` record (maybe a data source in terraform?)

lemeurherve: To be revised in another context than the migration one, WDYT?

dduportal: as seen together: first choice (merge, let the LB be created, retrieve its IP and add it in the DNS record )

dduportal (approved):

config/ldap_publick8s.yaml
lemeurherve: IP created in https://github.com/jenkins-infra/azure/pull/400

lemeurherve, line 2:
```suggestion
  type: LoadBalancer
  IP: 104.209.245.168
```

lemeurherve: Addressed in [`8db9381` (#4044)](https://github.com/jenkins-infra/kubernetes-management/pull/4044/commits/8db9381f9ae318998e6c6aca6ab2ab3acdd0221d)

lemeurherve:
hotfix: https://github.com/jenkins-infra/kubernetes-management/commit/98a46daa9a983ed8608007ca395b553c3be2cf32

PR #4101, opened 2023-06-22
Revert "feat(doks/doks-public) disabling during upgrade to 1.25"

smerle33:
Reverts jenkins-infra/kubernetes-management#4099

as per https://github.com/jenkins-infra/helpdesk/issues/3582

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -29,7 +29,7 @@ pipeline {
         axes {
           axis {
             name 'K8S_CLUSTER'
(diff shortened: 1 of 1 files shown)


timja (approved):

dduportal (approved):
- Only changes shown by the helmfile diff are the datadog patch update (#4098 was merged while doks* were disabled) as expected
- LGTM but we'll watch the potential warning during the helmfile apply step (to track api depreciations that we could have missed)

PR #4121, opened 2023-06-29
feat(javadoc): add arm64 affinity for javadoc

smerleCB:
not exclusive just soft affinity


dduportal:
As ssen together, it's not affitiny that we want in the current state of the infra, but tolerations (to allow scheduling on a tainted node).

PR #4152, opened 2023-07-07
Revert "disable AKS management for 1.25 upgrades"

dduportal:
Reverts jenkins-infra/kubernetes-management#4149


dduportal:
- No changes on both private and public clusters LGTM

PR #4160, opened 2023-07-10
hotifx(publick8s) put back javadoc workload to the arm64small nodepool

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3582#issuecomment-1629210833


dduportal:
Applied manually with success:

- Disabled the management job
- Scaled up the node pool manually to 1 node
- Applied the change

PR #4220, opened 2023-08-03
feat(publick8s) Force SSL redirect for jenkinsio

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3686

Forcing to SSL redirect is the default: https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#server-side-https-enforcement-through-redirect


dduportal:
Helmfil diff looks good (it removes the annotation as expected)

PR #4243, opened 2023-08-10
chore(accountapp): remove default `image.pullPolicy`

lemeurherve:
`IfNotPresent` is already the default value: https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

Supersedes and closes https://github.com/jenkins-infra/kubernetes-management/pull/4249

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -185,7 +185,7 @@ releases:
   - name: accountapp
     namespace: accountapp
     chart: jenkins-infra/accountapp
-    version: 0.5.26
+    version: 0.6.0
     needs:
       - ldap/ldap
     values:
--- a/config/accountapp.yaml
+++ b/config/accountapp.yaml
@@ -1,5 +1,3 @@
-image:
-  pullPolicy: IfNotPresent
 ingress:
   enabled: true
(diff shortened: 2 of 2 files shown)


dduportal (changes requested):
Alas it does not work as you would expect due to the chart being buggy: https://github.com/jenkins-infra/helm-charts/blob/2aff68e63375846a9942f8ff556d70b4a6e60f9c/charts/accountapp/templates/deployment.yaml#L29 

(The helmfile diff shows that the key is still specified but its value is undefined, which does not behave as described in the kube doc).


=> you might want to check the `helmfile diff` when doing such a change, before opening the PR to catch this kind of trick earlier

dduportal:
For info, blocked by https://github.com/jenkins-infra/helm-charts/pull/593 which should solve the issue in a proper way

lemeurherve:
Chart fixed in https://github.com/jenkins-infra/helm-charts/pull/593, released: https://github.com/jenkins-infra/helm-charts/releases/tag/accountapp-0.6.0

dduportal (approved):
LGTM, helmfile diff is 👍

PR #4244, opened 2023-08-10
chore: remove helmfile `--no-color` argument

lemeurherve:


dduportal (approved):
No objection as the ansicolor plugin is enabled in infra.ci

PR #4298, opened 2023-08-30
cleanup(ldap) remove legacy public IP of cert-ci from allow-list

dduportal:
Related to jenkins-infra/helpdesk#3688


dduportal (commented):
Helmfile diff looks good!

PR #4302, opened 2023-08-31
cleanup(ldap) remove IP of `vpn.jenkins.io` legacy VM from the allow …

dduportal:
…list

Ref https://github.com/jenkins-infra/helpdesk/issues/3689


dduportal:
Helmfile diff looks good 👍

PR #4334, opened 2023-09-11
feat(wiki): apply antiaffinity rule to avoid multiple pods on the same node

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3719

PR #4385, opened 2023-09-21
fix(wiki): change key selector for wiki antiaffinity

smerle33:
following https://github.com/jenkins-infra/kubernetes-management/pull/4334
correct the selector

### DIFF
--- a/config/wiki.yaml
+++ b/config/wiki.yaml
@@ -46,7 +46,7 @@ affinity:
     requiredDuringSchedulingIgnoredDuringExecution:
       - labelSelector:
           matchExpressions:
-            - key: app
+            - key: name
               operator: In
               values:
                 - wiki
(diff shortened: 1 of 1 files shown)


config/wiki.yaml
```diff
@@ -46,7 +46,7 @@ affinity:
     requiredDuringSchedulingIgnoredDuringExecution:
       - labelSelector:
           matchExpressions:
-            - key: app
+            - key: name
```
dduportal, line 49:
```suggestion
            - key: "app.kubernetes.io/name"
```

I believe we should use the full key name. Currently, the `metadata.labels` of each pod contains:

```
app.kubernetes.io/instance: wiki
app.kubernetes.io/name: wiki
```

dduportal (changes requested):

dduportal (approved):
LGTM! (helmfile diff looks good!)

PR #4395, opened 2023-09-22
chore(infra.ci.jenkins.io): move `helm-charts` job from `Kubernetes Management` folder to `Dependencies Management`

lemeurherve:
This PR moves this job which is only running updatecli.

We don't really care about build history for this job.

If it was the case, it would have needed to create the new job first, then connect into the controller and manually move the build history to this new job, then delete the old job.

### DIFF
--- a/config/jenkins-jobs_infra.ci.jenkins.io.yaml
+++ b/config/jenkins-jobs_infra.ci.jenkins.io.yaml
@@ -120,9 +120,6 @@ jobsDefinition:
             fileName: "kubeconfig"
             description: "Kubeconfig file for publick8s"
             secretBytes: "${base64:${KUBECONFIG_PUBLICK8S}}"
-      helm-charts:
-        name: Helm Charts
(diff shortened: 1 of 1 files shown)


dduportal (approved):
LGTM 👍 

Note: this will delete the build history, but we do not care for this job

dduportal:
Note: this caused unexpected wave of unwanted updatecli PRs. See https://github.com/jenkins-infra/helm-charts/pull/797 and https://github.com/jenkins-infra/kubernetes-management/pull/4422

hervelemeur:
@dduportal I've looked at the links but didn't see anything particular, how is it related to the updatecli wave?

dduportal:
> @dduportal I've looked at the links but didn't see anything particular, how is it related to the updatecli wave?

By moving (e.g. deleting + creating) without the build history, the new job rebuilt the last tags (the <3 days old) which opened way more PRs than expected.

PR #4430, opened 2023-09-26
feat(jenkinsio): add antiaffinity rule

smerle33:
as per jenkins-infra/helpdesk#3719 (comment)


dduportal (approved):
LGTM! We now have 2 different set of affinities for 2 different deployments 🥳 

Worked on this in pair with @smerle33 so he should be able to solve #4431 with the same method

PR #4432, opened 2023-09-26
feat(private-nginx-ingress_publick8s): add antiaffinity rule

smerle33:
as per jenkins-infra/helpdesk#3719 (comment)


dduportal (approved):
LGTM 👍  Diff looks good. Let's try this on the private ingress

PR #4540, opened 2023-10-16
feat(publick8s) migrate private-nginx-ingress to arm64 - controller and defaultBackend

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3619


dduportal (approved):
Diff is 👍 LGTM Let's deploy while we are watching

PR #4588, opened 2023-10-26
feat(doks, doks-public) disable clusters management

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3683#issuecomment-1780857936


dduportal:
Self merging as the 2 checks for the 2 DO clutsers are not run as expected

PR #4589, opened 2023-10-26
Revert "feat(doks, doks-public) disable clusters management (#4588)"

dduportal:
This reverts commit 631789bf01e289f275217dd70da81ed7ae502fa2.

It enables the management of the 2 clusters `doks` and `doks-public` after their upgrade to Kubernetes 1.26


dduportal (commented):
No helmfile diff found on both clusters: let's re-enable management

PR #4596, opened 2023-10-27
feat(publick8s): adding a service account for mirrorbits for update.jenkins.io

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/2649
provide a service account to trigger mirrobits scan for mirrors


dduportal (approved):
Helmfile diff is 👍

PR #4634, opened 2023-11-06
privatek8s: track JDK tool installations with `updatecli` + bump JDK8, JDK11 and JDK17 to latest available version

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3783, this PR adds updatecli manifests to track the JDK tool installation versions for JDK8, JDK11, JDK17 and JDK21.

To test the manifest, I've used it to update locally:

- JDK8 to `8u392-b08` (`updatecli apply` without `scmid` in targets)
- JDK11 to `11.0.16.1+1` (`updatecli apply` without `scmid` in targets)
- JDK17 to `17.0.8.1+1` ( Custom source + `updatecli apply` without `scmid` in targets) => this is the last version with x64_windows in the same version as Linux

Note: JDK21 is already up to date since #4629 

Note 2: https://github.com/updatecli/updatecli/issues/1733 would help to simplify the manifests here A LOT

PR #4635, opened 2023-11-06
disable AWS EKS clusters management

dduportal:
for https://github.com/jenkins-infra/helpdesk/issues/3683%5C#issuecomment-1794913444


dduportal:
Self-merging as the check is only due to updatecli (ref. https://github.com/jenkins-infra/kubernetes-management/pull/4634#issuecomment-1794976979)

PR #4640, opened 2023-11-07
Disable privatek8s

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3683#issuecomment-1797928819


dduportal:
Only 5 checks: let's roll!

PR #4663, opened 2023-11-14
feat(infra.ci) align Azure-VM agent sizing with ci.jenkins.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3812

This PR ensures we use the same kind of instance (lower costs for spots) with ephemeral disks (faster startup and no $$ to pay for managed disks).

Note that our quota might be limiting : let's keep it like this for now until we start using the new MSOP subscription.


smerle33 (approved):
🚀

PR #4742, opened 2023-12-05
feat(publick8s): adding a toleration for system nodepool on datadog and falco

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3827
specify a toleration for datadog and falco in order to get spawn on system nodepool after we tainted it

### DIFF
--- a/config/datadog_publick8s.yaml
+++ b/config/datadog_publick8s.yaml
@@ -14,6 +14,10 @@ agents:
       operator: "Equal"
       value: "arm64"
       effect: "NoSchedule"
+    - key: "kubernetes.azure.com/mode"
+      operator: "Equal"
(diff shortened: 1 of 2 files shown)


config/datadog_publick8s.yaml
```diff
@@ -22,3 +26,7 @@ clusterAgent:
       operator: "Equal"
       value: "arm64"
       effect: "NoSchedule"
+    - key: "kubernetes.azure.com/mode"
+      operator: "Equal"
+      value: "system"
+      effect: "NoSchedule"
```
dduportal, line 32:
```suggestion
```

We don't want clusteragent to be allowed to run on a x86 machine neither on the system pool

PR #4743, opened 2023-12-05
feat(publick8s): relaxing node selection to amd64 instead of specific nodepool

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3827


dduportal (approved):
🚀

PR #4758, opened 2023-12-11
feat(infra.ci): add a child to reports

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3832

provide an infra.ci.jenkins.io child to reports for mirrorbits mirror list for downloads.

### DIFF
--- a/config/jenkins-jobs_infra.ci.jenkins.io.yaml
+++ b/config/jenkins-jobs_infra.ci.jenkins.io.yaml
@@ -219,10 +219,10 @@ jobsDefinition:
             username: "${JIRA_API_INFRA_REPORTS_USERNAME}"
             usernameSecret: true
             password: "${JIRA_API_INFRA_REPORTS_PASSWORD}"
-      mirrorbits-mirrors-list:
-        name: "List of Mirrorbits mirrors"
+      jenkins-infra-data:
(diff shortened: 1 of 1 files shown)


config/jenkins-jobs_infra.ci.jenkins.io.yaml
```diff
@@ -219,6 +219,10 @@ jobsDefinition:
             username: "${JIRA_API_INFRA_REPORTS_USERNAME}"
             usernameSecret: true
             password: "${JIRA_API_INFRA_REPORTS_PASSWORD}"
+      mirrorbits-mirrors-list:
+        name: "List of Mirrorbits mirrors"
+        repository: "infra-reports"
+        jenkinsfilePath: "mirrorbits-mirrors-list/Jenkinsfile"
```
dduportal, line 225: WDYT about the following naming which extends the scope from "mirrorbits mirror list" to "Jenkins Infrastructure public data" (as we will be able to add our outbound and inbound IPs to allow user firewall to be set up with automation for example) ?

```suggestion
      jenkins-infra-data:
        name: "Jenkins Infrastructure Public Data"
        repository: "infra-reports"
        jenkinsfilePath: "jenkins-infra-data/Jenkinsfile"
```

dduportal (approved):

PR #4847, opened 2024-01-08
feat(publick8s) install a new instance of get.jenkins.io with mirrorb…

dduportal:
…its-parent chart


dduportal:
- Added a second (distinct - easier to revert later) commit to restrict the new instance to only the `azure.get.jenkins.io` domain and avoid polluting existing service: https://github.com/jenkins-infra/kubernetes-management/pull/4847/commits/56521c2aa8fe780308add0c8ef432fd52def2543
- Same idea in the secrets: let's use a distinct RedisDb (ID `1` instead of `0`) for the initial installation - https://github.com/jenkins-infra/charts-secrets/commit/4ee960ea92afb1b2f9bf57a8771d41fca71ee73e

dduportal:
Installation done manually "by part" as the bootstrap failed from scratch (worked on my test cluster but not in production...).

- Started with the global file storage
- httpd then
- mirrorbit-list then
- Fixed the missing parameters (replica count, etc.)
- Then ingress

=> test is successul with my `/etc/hosts` set up for the domainb `azure.get.jenkins.io` to point to the cluster IP.

Let's merge this first instance

PR #4909, opened 2024-01-26
WIP(do not merge) configure a temporary instance of get-jenkins-io for helpdesk-3917

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3917

PR #4911, opened 2024-01-26
feat(get.jenkins.io) Migrate to the new premium storage account

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3917

Follow up of manual tests done in #4909 

This PR **needs** the secrets to be updated. As such, the deployment will be run manually.


dduportal (commented):
The (visible) changelog shows:

- Volumes mounted in readonly on `mirrorbits` pods (expected)
- Change of RG name for the data volume


Let's roll

PR #4939, opened 2024-02-05
chore(updatecli) fix all jenkins manifests related to 5.x chart's image specification

dduportal:
Fixup of #4925 

Closes #4934 closes #4933 
(new PRs will be opened)


dduportal:
Self merging as we want only updatecli

PR #4999, opened 2024-02-20
fix(uplink) changing replicaCount to 1 as it's not able to use multiple pods

smerle33:
https://github.com/jenkins-infra/helpdesk/issues/3883


dduportal (approved):
As this application does not fit to be scaled (no shared lock)

PR #5023, opened 2024-02-26
chore(jenkins.io): migrate to its new `jenkinsio` storage account

lemeurherve:
This PR uses the last version 1.2.0 of jenkinsio chart to allow specifying a different storage account for jenkinsio & jenkinsio-zh so we can migrate jenkins.io to its new `jenkinsio` storage account.

I've put the original `prodjenkinsio` storage account credentials as a `zh` one in https://github.com/jenkins-infra/charts-secrets/blob/main/config/jenkinsio/secrets2.yaml (private repo) alongside the new `jenkinsio` storage account (which does not contain the `zhjenkinsio` File Share, only the new `jenkins-io` one) to avoid service disruption.
I'll update https://github.com/jenkins-infra/charts-secrets/blob/main/config/jenkinsio/secrets.yaml to contain both storage account credentials and change the release config in a follow-up pull request.

Tested on another release locally with success.

Supersedes and closes #5022

Follow-up of:
- https://github.com/jenkins-infra/jenkins.io/pull/7123

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414

PR #5024, opened 2024-02-26
chore(jenkins.io): use secrets.yaml instead of secrets2.yaml

lemeurherve:
This PR is the follow-up to reuse secrets.yaml now updated with the content of secrets2.yaml. (Two steps migration to avoid service disruption)

Follow-up of:
- #5023 

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414


lemeurherve:
Diffs LGTM, self-merging to continue on the subject.

PR #5025, opened 2024-02-26
chore(infra-jobs): remove `githubCheckName`s

lemeurherve:
This PR removes the GitHub check names, now useless since the use of https://github.com/jenkinsci/github-scm-trait-notification-context-plugin:

image

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2778#issuecomment-1959565465

### DIFF
--- a/config/jenkins-jobs_infra.ci.jenkins.io.yaml
+++ b/config/jenkins-jobs_infra.ci.jenkins.io.yaml
@@ -145,14 +145,12 @@ jobsDefinition:
         name: Puppet (jenkins-infra)
         jenkinsfilePath: Jenkinsfile_updatecli
         enableGitHubChecks: true
-        githubCheckName: "Updatecli (infra.ci.jenkins.io)"
         branchIncludes: "production staging updatecli_* PR-* main"
         disableTagDiscovery: true
       kubernetes-management:
         jenkinsfilePath: Jenkinsfile_updatecli
         disableTagDiscovery: true
         enableGitHubChecks: true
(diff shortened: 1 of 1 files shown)


timja (approved):
FWIW those status checks names are incredible long =/

lemeurherve:
> FWIW those status checks names are incredible long =/

I'd suggest removing "jenkins" prefix (does not add meaningful information), the job name (as there isn't any case where two jobs are running in the same folder AFAIK, and as it can easily be read from the URL on link hover), and maybe the "pr-merge/branch" suffix.


From:
```
jenkins/infra.ci.jenkins.io/kubernetes-jobs/kubernetes-management/pr-merge
jenkins/infra.ci.jenkins.io/updatecli/kubernetes-management/pr-merge
```

To:
```
infra.ci.jenkins.io/kubernetes-jobs
infra.ci.jenkins.io/updatecli
```

WDYT @timja @dduportal @smerle33?

timja:
I would think about what is important.

to me that's the check name and it should be first.
There's no need to use / to make it harder to read.

to me its:
`kubernetes-management (infra.ci)` or even `jenkins` or `build` for that folder.

do you even need the jenkins instance for updatecli?
`updatecli` or `updatecli (infra.ci)`

it'll be in the actor's name of who creates the check anyway

lemeurherve:
Agree, the thing is there are some constraints to get it automatically named AFAIU cf @dduportal's explanations at https://github.com/jenkins-infra/helm-charts/pull/1081

PR #5027, opened 2024-02-27
chore(plugins.jenkins.io): use new `pluginsjenkinsio` storage account

lemeurherve:
This PR  migrates plugins.jenkins.io to its new `pluginsjenkinsio` storage account, using an intermediate secret to avoid any service interruption.

Needs:
- https://github.com/jenkins-infra/plugin-site/pull/1596

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-1964568638

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -215,7 +215,7 @@ releases:
     values:
       - "../config/plugin-site.yaml"
     secrets:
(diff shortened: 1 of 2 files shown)


lemeurherve:
Diff LGTM, status opened in case of issue (there shouldn't be any), self-merging to continue on the subject.

PR #5028, opened 2024-02-28
chore(plugins.jenkins.io): use secrets.yaml instead of intermediate-secrets.yaml

lemeurherve:
This PR get rid of the intermediate secret used to avoid service interruption while migrating to the new storage account.

Follow-up of:
- #5027 

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-1964568638


lemeurherve:
Self-merging to continue on the subject.

PR #5029, opened 2024-02-28
chore(javadoc.jenkins.io): use new `javadocjenkinsio` storage account

lemeurherve:
This PR  migrates javadoc.jenkins.io to its new `javadocjenkinsio` storage account, using an intermediate secret to avoid any service interruption.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-1966067802


lemeurherve:
Diff LGTM, self-merging to continue on the subject.

PR #5088, opened 2024-03-25
chore(updates.jenkins.io): activate HA for mirrorbits and httpd

lemeurherve:
This PR set the replica count to 2 for the mirrorbits and the httpd subcharts of updates.jenkins.io to activate the High Availability of this service.

As the rsync server won't be requested as much as mirrorbits and Apache, I've kept its replica count to 1.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649

### DIFF
--- a/config/updates.jenkins.io.yaml
+++ b/config/updates.jenkins.io.yaml
@@ -45,6 +45,7 @@ global:
 
 mirrorbits:
   enabled: true
+  replicaCount: 2
   resources:
     limits:
       cpu: 2
@@ -57,6 +58,7 @@ mirrorbits:
 
 httpd:
   enabled: true
+  replicaCount: 2
   resources:
     limits:
       cpu: 1000m


dduportal (approved):
LGTM!

2 nitpicks (NOT blocking!)

1. I would change the title:
  - This change is  anew feature as it changes something for end users: requests are spread across replicas and the availability is increased
  - The "component" should be `publick8s` or eventually `publick8s/updates.jenkins.io` to help the reviewer (or us in 3 months) to immediately locate which cluster was impacted. This one is a nit because we do not have a formal convention.

2. The helmfile diff only shows a Pod Disruption Budget for `mirrorbits` but not for the `httpd` component: worth checking once merged and deployed successfully (maybe it already exists, or maybe the httpd chart does not define a PDB when replicating). Might be worth checking along with get.jenkins.io

dduportal:
> LGTM!
> 
> 2 nitpicks (NOT blocking!)
> 
>     1. I would change the title:
> 
> 
>     * This change is  anew feature as it changes something for end users: requests are spread across replicas and the availability is increased
> 
>     * The "component" should be `publick8s` or eventually `publick8s/updates.jenkins.io` to help the reviewer (or us in 3 months) to immediately locate which cluster was impacted. This one is a nit because we do not have a formal convention.
> 
> 
>     2. The helmfile diff only shows a Pod Disruption Budget for `mirrorbits` but not for the `httpd` component: worth checking once merged and deployed successfully (maybe it already exists, or maybe the httpd chart does not define a PDB when replicating). Might be worth checking along with get.jenkins.io

Issue: https://github.com/jenkins-infra/helm-charts/issues/1109

PR #5118, opened 2024-04-09
cleanup(publick8s/plugin-site-issues) rollback custom annotations back to default key algorithm (RSA)

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3978

As there are no strong compelling reason to use ECDSA over RSA and we can't default to ECDSA by default, this PR rolls back the changes introduced in https://github.com/jenkins-infra/kubernetes-management/pull/5080 and https://github.com/jenkins-infra/kubernetes-management/pull/5110 as we want to keep with the same default settings for all of our certificates: enforcing the ECDSA-related annotation across all ingresses is time-consuming.

Please note that a one-time manual change will be required to re-issue an RSA certificate by adding the following annotation to the `plugin-site-issue` ingress right before force-renewing (with `cmctl`):

```
cert-manager.io/private-key-rotation-policy: Always
```

### DIFF
--- a/config/plugin-site-issues.yaml
+++ b/config/plugin-site-issues.yaml
@@ -3,8 +3,6 @@ ingress:
   className: public-nginx
   annotations:
     "cert-manager.io/cluster-issuer": "letsencrypt-prod"
-    "cert-manager.io/private-key-algorithm": "ECDSA"
-    "cert-manager.io/private-key-rotation-policy": "Always"
     "nginx.ingress.kubernetes.io/ssl-redirect": "true"
     "nginx.ingress.kubernetes.io/enable-cors": "true"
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

smerle33 (approved):
🚀

dduportal:
- Deployed with success: annotations removed
- Regenerated an RSA certificate with success:
    - Added back the annotation to always renew the private key when the key encryption algorithm is changed
    - Ran the `cmctl renew -n plugin-site plugin-site-issues-tls`
    - After 2 min, the `kubectl -n plugin-site describe certificate plugin-site-issues-tls` showed the new certificate to be installed
    - Private browsing to plugin site shows an RSA cert

PR #5126, opened 2024-04-12
chore(infra.ci): migrate to arm64

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3823#issuecomment-2036372624

this will migrate infra.ci to it's specific pool which is arm64 as per https://github.com/jenkins-infra/azure/pull/658 

----
Deployment procedure

as this is an heavy change on infra.ci infrastructure it is safer to apply it manually (instead of relying on agent re-connection if everything goes according to the plan) 

- from a local administrator machine with access to privatek8s API (and same kubectl / helmfile version as packer image allinone agent)
- ensure that the local kubernetes-management repository is up to date with this PR (including `./secrets` sub repository)
- execute helmfile diff to ensure that the change is correctly plan
- deploy with helmfile apply
    - merge this PR as soon as the pod is scheduled and before infra.ci is up and running

PR #5132, opened 2024-04-12
fix(privatek8s/datadog) ensure datadog agent runs on all nodes by updating tolerations

dduportal:
Follow up of #5126
Related to https://github.com/jenkins-infra/helpdesk/issues/3823

This PR fixes the datadog configuration in the `privatek8s` cluster to ensure we have an agent per node.

Today we only have datadog agents on the syspool and linuxpool node pools.
The arm64 node pools have taints forbidding the datadog agent to be scheduled.

By adding tolerations to the datadog deployment, we allow them to be scheduled on any kind of nodes.

----

Notes:

- The Windows node pool (used only during Jenkins core releases by release.ci.jenkins.io is out of scope)
- The tolerations added in this PR covers ALL the other node pools: all controllers, all agents, etc.
- Tested manually with success on production and then roll backed.
- Thanks @MarionPlt for the rubber ducking on this one!

### DIFF
--- a/config/datadog_privatek8s.yaml
+++ b/config/datadog_privatek8s.yaml
@@ -8,3 +8,29 @@ datadog:
     hostCAPath: /etc/kubernetes/certs/kubeletserver.crt
     # Required as of Agent 7.35 because Kubelet certificates in AKS do not have a Subject Alternative Name (SAN) set.
     tlsVerify: false
+agents:
+  tolerations:
+    - key: "kubernetes.io/arch"
+      operator: "Equal"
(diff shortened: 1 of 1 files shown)


dduportal (commented):
Self-merging:
- Tested in production with success
- Required to ensure infra.ci in arm64 continues working
- We need METRICS

Note: the helmfile diff shows (on privatek8s) a cleanup of changes incoming on a subsequent PR (DD_HOSTNAME to support Azure Linux)

PR #5175, opened 2024-04-25
feat(updates.jenkins.io): set azure.updates.jenkins.io backend to `httpd` and mirrors.updates.jenkins.io to `mirrorbits`

lemeurherve:
This PR update the new Update Center deployment (Azure + mirrorbits) to have the described setup (azure.updates.jenkins.io => httpd and mirrors.updates.jenkins.io => mirrorbits).

As we don't need to filter path with nginx-ingress & regexp anymore, removing the correspond snippet annotation.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2075477428


lemeurherve:
Helmfile diff LGTM, self-merging to continue on the subject.

PR #5184, opened 2024-04-30
feat(updates.jenkins.io): dedicated File Share for the `httpd` service

lemeurherve:
As the httpd service doesn't need to have a common File Share with the mirrorbits service, this PR ~disables it from the mirrorbits-parent release~ (extracted in #5186), and adds a dedicated httpd release from its own chart so it can use its own updates-jenkins-io-httpd File Share created in https://github.com/jenkins-infra/azure/pull/676.

We're keeping the mirrorbits and the rsync services under mirrorbits-parent umbrella as they still need to share the same updates-jenkins-io File Share.

Corresponding secret added in https://github.com/jenkins-infra/charts-secrets/commit/688566ee492e9026f09289d79ece7f8f4f06cdf3 (private repo).

Follow-up of:
- https://github.com/jenkins-infra/kubernetes-management/pull/5175

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2076659536

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -241,6 +241,14 @@ releases:
       - "../config/updates.jenkins.io.yaml"
     secrets:
       - "../secrets/config/updates.jenkins.io/secrets.yaml"
+  - name: updates-jenkins-io-httpd
+    namespace: updates-jenkins-io
+    chart: jenkins-infra/httpd
+    version: 0.4.0
+    values:
+      - "../config/updates.jenkins.io_httpd.yaml"
+    secrets:
+      - "../secrets/config/updates.jenkins.io/httpd-secrets.yaml"
   - name: contributors-jenkins-io
     namespace: contributors-jenkins-io
     chart: jenkins-infra/nginx-website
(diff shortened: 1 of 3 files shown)


lemeurherve:
I'll split this PR in two separate ones to avoid the current error:

> Error: Failed to render chart: exit status 1: Error: Unable to continue with install: ConfigMap "updates-jenkins-io-httpd" in namespace "updates-jenkins-io" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "updates-jenkins-io-httpd": current value is "updates-jenkins-io"

lemeurherve:
Opened https://github.com/jenkins-infra/kubernetes-management/pull/5186 to be merged before this PR.

dduportal (commented):
LGTM! The helmfile diff looks really good. You can merge whenever you are ready to monitor deployment!

lemeurherve:
Bummer...

> PersistentVolume "httpd-binary" is invalid: [spec.capacity: Required value, spec.capacity: Unsupported value: core.ResourceList(nil): supported values: "storage", spec.azureFile.secretName: Required value]

Reverted, will rework it later.

PR #5190, opened 2024-04-30
Revert "feat(updates.jenkins.io): dedicated File Share for the `httpd` service"

lemeurherve:
Reverts jenkins-infra/kubernetes-management#5184

From build on main:

> PersistentVolume "httpd-binary" is invalid: [spec.capacity: Required value, spec.capacity: Unsupported value: core.ResourceList(nil): supported values: "storage", spec.azureFile.secretName: Required value]

Will rework it later.


lemeurherve:
Self-merging to restore builds on main.

PR #5192, opened 2024-05-02
feat(publick8s/updates.jenkins.io): deploys a dedicated `httpd` service with a distinct fileshare

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2089993133.

The goal of this PR is to instantiate an `httpd` service for the `azure.updates.jenkins.io` webservice with its own distinct fileshare.

It is a second tentative of delivering #5184 with an additional fixup commit to ensure PV + PVC are created without error and it reverts #5190.

The `httpd` chart does not require (at least for now) a change: we can set up a statically provisioned PVC.

Validations:
- The official AKS documentation was used to set up the attributes: https://learn.microsoft.com/en-us/azure/aks/azure-csi-files-storage-provision#statically-provision-a-volume
- The resulting YAML had been statically verified: 
  - The command `helmfile template -f ./clusters/publick8s.yaml -l name=updates-jenkins-io-httpd > .tmp/uc-httpd.yaml` has been used to generate a local Kubernetes manifest file for ALL objects
  - This file has been linted and validated with the official Kubernetes JSON schemas to ensure all attributes are 👍 
- Then, the triplet secret + PV + PVC had been provisionned in the `default` namespace of `publick8s` using the (stripped from other objects such as `Deployment` file `tmp/uc-httpd.yaml`) with success: the PVC was seen as a 1 Gb  Bound volume!).

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -241,6 +241,14 @@ releases:
       - "../config/updates.jenkins.io.yaml"
     secrets:
       - "../secrets/config/updates.jenkins.io/secrets.yaml"
+  - name: updates-jenkins-io-httpd
+    namespace: updates-jenkins-io
+    chart: jenkins-infra/httpd
+    version: 0.4.0
+    values:
+      - "../config/updates.jenkins.io_httpd.yaml"
+    secrets:
+      - "../secrets/config/updates.jenkins.io/httpd-secrets.yaml"
   - name: contributors-jenkins-io
     namespace: contributors-jenkins-io
     chart: jenkins-infra/nginx-website
--- a/config/updates.jenkins.io_httpd.yaml
+++ b/config/updates.jenkins.io_httpd.yaml
@@ -14,42 +14,58 @@ tolerations:
     value: "arm64"
     effect: "NoSchedule"
 repository:
-  name: httpd-binary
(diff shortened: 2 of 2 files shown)


dduportal:
For info:

- This PR failed to deploy. The `helmfile` release failed to deploy with a `Context deadline exceeded` error meaning there was a (many?) Kubernetes object not able to be created properly. ⚔️ 
- Secret, PV and PVC were created with success and PVC was marked as Bound ✅ 
- But the 2 pods were stuck at the `ContainerCreating` phase. A `describe pod` command showed the following error message:

  ```
   MountVolume.MountDevice failed for volume "updates-jenkins-io-httpd" : rpc error: code = InvalidArgument desc = GetAccountInfo(updates-jenkins-io-httpd) failed with error: could not find accountname or azurestorageaccountname field in secrets
  ```
- The Secret name is properly used by the PV. But the content of the secret is not structured as expected. We updated the secret `data` keys for mirrorbits-parent chart to support CSI, but we never cherry-picked this for `httpd` secrets.


- Problem fixed by @lemeurherve by updating the chart secrets in (private link) https://github.com/jenkins-infra/charts-secrets/commit/634e3502e6f228b329e61517bc45f77e3d31a18d

PR #5312, opened 2024-06-17
feat(privatek8s/infra.ci.jenkins.io): add `contributor-spotlight` updatecli job

lemeurherve:
Ref:
- https://github.com/jenkins-infra/contributor-spotlight/issues/124
- https://github.com/jenkins-infra/contributor-spotlight/pull/146


config/jenkins-jobs_infra.ci.jenkins.io.yaml
```diff
@@ -147,6 +147,8 @@ jobsDefinition:
       infra-reports:
         jenkinsfilePath: Jenkinsfile_updatecli
         disableTagDiscovery: true
+      contributor-spotlight:
+        disableTagDiscovery: true
```
dduportal, line 151: This setup uses `Jenkinsfile` as default. But we don't want to rebuild the whole [`Jenkinsfile`](https://github.com/jenkins-infra/contributor-spotlight/blob/main/Jenkinsfile) right?

As such, I propose you to use the same pattern as other updatecli jobs:

```suggestion
      contributor-spotlight:
        jenkinsfilePath: Jenkinsfile_updatecli
        disableTagDiscovery: true
```

THen, once merged, no build will be run for this new job until you add a `Jenkinsfile_updatecli` in a PR (like... https://github.com/jenkins-infra/contributor-spotlight/pull/146 :) )

dduportal (changes requested):
Waiting for a decision on which pipeline to use, but we cannot merge as it otherwise we would compete with the standard job

dduportal:
Nit: PR title should be changed to something like `feat(privatek8s/infra.ci.jenkins) add a new updatecli job for contributor-spotlight`

lemeurherve:
Reformulate the title like the one you renamed earlier.

(Note: I would have skipped "privatek8s", kinda irrelevant to know which cluster is hosting infra.ci, the important bit is on which instance this job is IMO)

dduportal:
> (Note: I would have skipped "privatek8s", kinda irrelevant to know which cluster is hosting infra.ci, the important bit is on which instance this job is IMO)

I strongly disagree with this stance (irrelevance):
- You cannot guess where is `infra.ci.jenkins.io` is set up unless you are a hardcore maintainer of the repository. 
- Specifying the cluster in the scope helps to visualize which cluster is the entrypoint and what could be the impact of a given PR (which is kinda required for a production setup)

I agree that it is also important to have `infra.ci.jenkins.io` in the scope because it helps scoping even more

lemeurherve:
Apart from this disagreement on a note, is this PR OK to you?

config/jenkins-jobs_infra.ci.jenkins.io.yaml
lemeurherve: Thanks!

dduportal (approved):

PR #5317, opened 2024-06-17
feat(privatek8s/infra.ci.jenkins.io): add `stats.jenkins.io` updatecli job

lemeurherve:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4132#issuecomment-2167717401
- https://github.com/jenkins-infra/stats.jenkins.io/pull/5


config/jenkins-jobs_infra.ci.jenkins.io.yaml
```diff
@@ -151,6 +151,8 @@ jobsDefinition:
       infra-reports:
         jenkinsfilePath: Jenkinsfile_updatecli
         disableTagDiscovery: true
+      stats.jenkins.io:
+        disableTagDiscovery: true
```
dduportal, line 155: This setup uses `Jenkinsfile` as default. But we don't want to rebuild the whole [`Jenkinsfile`](https://github.com/jenkins-infra/stats.jenkins.io/blob/main/Jenkinsfile).

As such, I propose you to use the same pattern as other updatecli jobs:
```suggestion
      stats.jenkins.io:
        jenkinsfilePath: Jenkinsfile_updatecli
        disableTagDiscovery: true
```

Then, once merged, no build will be run for this new job until you add a `Jenkinsfile_updatecli` in a PR of https://github.com/jenkins-infra/stats.jenkins.io

dduportal (changes requested):
Waiting for a decision on which pipeline to use, but we cannot merge as it otherwise we would compete with the standard job

dduportal:
Nit: PR title should be changed to ` feat(privatek8s/infra.ci.jenkins) add a new updatecli job for stats.jenkins.io`

dduportal (approved):

PR #5337, opened 2024-06-21
feat(infracijioagents1): removing temporarily from management to upgrade to kubernetes 1.28

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4144#issuecomment-2178169723

we remove the cluster `infracijioagents1`from the kubenetes-management to handle is upgrade to kubernetes 1.28


jayfranco999 (approved):
LGTM, let's go ahead with the upgrade 🚀

PR #5342, opened 2024-06-21
feat(privatek8s/infra.ci.jenkins.io) migrate container agents workload to infracijioagents1

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3923

This PR migrates the container agents workloads of infra.ci.jenkins.io to the new `infracijio-agent-1` cluster

Note: 

- The agent pod template `jnlp-windows` is removed as we do not have Windows Node pool on the new cluster (and it is not used: we use Windows Azure VM if needed)
- The agent pod template `jnlp-linux` is renamed to `jnlp-linux-amd64`. It keeps the same labels as before + addition label `jnlp-linux-amd64`
- The env. var `ARTIFACT_CACHING_PROXY_PROVIDER` is added to `jnlp-linux-amd64` and `jnlp-linux-arm64`

### DIFF
--- a/config/jenkins_infra.ci.jenkins.io.yaml
+++ b/config/jenkins_infra.ci.jenkins.io.yaml
@@ -126,24 +126,20 @@ controller:
           clouds:
             - kubernetes:
                 containerCapStr: "100"
-                # Controller URL to allow agent to connect to it. Prefers using the Kubernetes Service as an internal URL.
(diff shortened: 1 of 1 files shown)


dduportal:
Self-merging:

- Announced on IRC
- Helmfile diff looks good

PR #5348, opened 2024-06-24
feat(cijioagents1): removing temporarily from management to upgrade to Kubernetes 1.28

jayfranco999:
Removing cjioagents1 cluster from kub management to proceed with kub 1.28 upgrade.


smerle33 (approved):
🚀

PR #5356, opened 2024-06-26
chore(privatek8s/infra.ci.jenkins.io): new websites File Share service principal writer credential names

lemeurherve:
This PR renames website File Share service principal writer credentials so they're the same than the ones used in https://github.com/jenkins-infra/azure/blob/91b75632186a6ab529a858166de84456bf1e87dd/updatecli/values.yaml to ensure consistency and avoid mistakes.

EDIT: duplicating them as a first step, will remove duplicated credentials after updating consumers pipelines.

Follow-up of:
- https://github.com/jenkins-infra/azure/pull/754

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4149

### DIFF
--- a/config/jenkins-jobs_infra.ci.jenkins.io.yaml
+++ b/config/jenkins-jobs_infra.ci.jenkins.io.yaml
@@ -441,7 +441,7 @@ jobsDefinition:
           algolia-plugins-write-key:
             secret: "${ALGOLIA_PLUGINS_JENKINS_IO_WRITE_KEY}"
             description: "Algolia credentials to write data for plugin site"
-          # TODO: remove after credentials renaming in contributor-spotlight pipeline
+          # TODO: remove after credentials renaming in plugin-site pipeline
           infraci-pluginsjenkinsio-fileshare-service-principal-writer:
             kind: azure-serviceprincipal
             azureEnvironmentName: "Azure"
@@ -516,7 +516,7 @@ jobsDefinition:
         jenkinsfilePath: Jenkinsfile
         enableGitHubChecks: true
         credentials:
-          # TODO: remove after credentials renaming in contributor-spotlight pipeline
+          # TODO: remove after credentials renaming in docs.jenkins.io pipeline
           infraci-docs-jenkins-io-fileshare-service-principal-writer:
             kind: azure-serviceprincipal
             azureEnvironmentName: "Azure"
@@ -552,7 +552,7 @@ jobsDefinition:
         branchIncludes: main
         disableTagDiscovery: true
         credentials:
-          # TODO: remove after credentials renaming in contributor-spotlight pipeline
+          # TODO: remove after credentials renaming in stats.jenkins.io pipeline
           infraci-stats-jenkins-io-fileshare-service-principal-writer:
             kind: azure-serviceprincipal
             azureEnvironmentName: "Azure"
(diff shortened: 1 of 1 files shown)


config/jenkins-jobs_infra.ci.jenkins.io.yaml
```diff
@@ -441,7 +441,7 @@ jobsDefinition:
           algolia-plugins-write-key:
             secret: "${ALGOLIA_PLUGINS_JENKINS_IO_WRITE_KEY}"
             description: "Algolia credentials to write data for plugin site"
-          infraci-pluginsjenkinsio-fileshare-service-principal-writer:
+          infraci_pluginsjenkinsio_fileshare_serviceprincipal_writer:
```
dduportal, line 444: - This YAML key is not using the same convention as the other keys (above and below) with this change, which defeats the purpose of the change itself (avoid confusion)
- Changing the key id means breaking the associated pipeline (ref. https://github.com/jenkins-infra/plugin-site/blob/7d3ce54af398b75553e96cdbc859fa0b0960f4a6/Jenkinsfile#L132) until it is updated.


=> As much as the other renaming make sense in the context of terraform itself, this change here is not wanted unless a compelling reason

dduportal (changes requested):

config/jenkins-jobs_infra.ci.jenkins.io.yaml
lemeurherve:
```suggestion
          infraci-pluginsjenkinsio-fileshare-service-principal-writer:
            kind: azure-serviceprincipal
            azureEnvironmentName: "Azure"
            clientId: "${INFRACI_PLUGINSJENKINSIO_FILESHARE_SERVICE_PRINCIPAL_WRITER_CLIENT_ID}"
            clientSecret: "${INFRACI_PLUGINSJENKINSIO_FILESHARE_SERVICE_PRINCIPAL_WRITER_PASSWORD}"
            description: "plugins.jenkins.io File Share Service Principal Writer"
            subscriptionId: "${JENKINSINFRA_AZURE_PRIMARY_SUBSCRIPTION_ID}"
            tenant: "${JENKINSINFRA_AZURE_TENANT_ID}"
          infraci_pluginsjenkinsio_fileshare_serviceprincipal_writer:
```
Does this change duplicating the credentials until the plugin-site pipeline is updated would be acceptable?

Having the same name format everywhere would be nice for consistency.

lemeurherve: All credentials were concerned, hence https://github.com/jenkins-infra/kubernetes-management/pull/5356/commits/bebacc4caba04281edf30560d65c20ec47892783

WDYT?

dduportal: > Does this change duplicating the credentials with a new name until the plugin-site pipeline is updated would be acceptable?

In term of "not breaking the pipeline": yes

> Having the same name format everywhere would be nice for consistency.

In theory: yes. In practice: 
- Not if it create inconsistencies everywhere else (like this PR unless we change ALL credentials inb this file with the associated effort to update all pipeline) => not worth the effort
- And not if the convention become a hard process constraining everything: if terraform objects naming convention constrains the Jenkins credential IDs specified by the Jenkins pipeline (unrelated areas, and potentially unrelated actors) => coupling is the worst we could make in an infra-as-code world.

PR #5363, opened 2024-06-27
feat(privatek8s/infra.ci.jenkins.io) bump image version to `1.48.1-2.464`

dduportal:
Ref. https://www.jenkins.io/security/advisory/2024-06-26/

Manual apply as there are not automation


smerle33 (approved):
🚀

PR #5388, opened 2024-07-04
feat(privatek8s/infra.ci): migrate volume/pv/pvc to standard ZRS

dduportal:
as per https://github.com/jenkins-infra/helpdesk/issues/4044
migrating to terraform managed disk/pv/pvc in standard ZRS

⚠️ do not merge, need a last moment sync for the migration


smerle33 (approved):
🚀

PR #5394, opened 2024-07-05
feat(publick8s) decrease resource requests to pack more pods

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/3837, we saw in datadog that the "real life" usage of most of the arm64 services is way below the resource requests.

This PR optimizes this to ensure we can pack more per nodes on arm64


dduportal (commented):
Self merging

PR #5510, opened 2024-08-06
fix(publick8s/public-ingress) ensure external LBs are SingleStack to prevent errors trying to create unused IPv6s

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4206

This PR fixes the 2 errors related to `Authorization Failures` with IPv6 Public IP deprecated and being reused (trying to) by the LB.

The root causes are the same: using DualStack LBs instead of 2 SingleStack LBs.

- Fixing the Nginx ingress LB  using helm chart values specifying the SingleStack `IPv4`
- Fixing the custom IPv6 LB using its new `1.0.1` (fixed) chart from https://github.com/jenkins-infra/helm-charts/pull/1251

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -223,7 +223,7 @@ releases:
   - name: ipv6-lb-service
     namespace: public-nginx-ingress
     chart: jenkins-infra/ipv6-lb-service
-    version: 1.0.0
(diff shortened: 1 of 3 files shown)


dduportal:
Merging as it was applied manually

PR #5534, opened 2024-08-12
fix(publick8s/mirrors.updates.jio) redirect the app-root to `/index.html`

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2283801444

This PR fixes the "too much redirection" when hitting the app root (https://azure.updates.jenkins.io -> https://mirrors.updates.jenkins.io -> https://azure.updates.jenkins.io etc.).

It set up the ingress to immediately redirects any exact request on `/` to the `/index.html` file explicitly so the file is served by a mirror (instead of the Azure cluster).


dduportal (commented):
Self merging: tested manually with success and helmfile diff is looking good

PR #5566, opened 2024-08-17
cleanup(publick8s/privatek8s) remove leftovers of the public ACP instance

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4204

This PR removes all references (in this repo) to the "public" ACP which we want to deprovision:

- Helmfile tracking in publick8s
- Fix the updatecli ACP-related manifests
- Stop using ACP in infra.ci.jenkins.io


dduportal:
Helmfile diff looks good 👍 Let's merge, and then we'll remove the NS

PR #5571, opened 2024-08-17
feat(publick8s) enable datadog logs collection on mirrorbits

dduportal:


dduportal (commented):
Let's roll and avoid filling nodes disks!

PR #5579, opened 2024-08-19
fix(publick8s) ensure enforced security for both get.jenkins.io and azure.updates.jenkins.io

dduportal:

Requires the version 5.0.1 of the mirrorbits chart (deployed with https://github.com/jenkins-infra/kubernetes-management/pull/5578)

This PR ensures that both mirrorbits systems have enforced security:

- No privilege escalation possible
- No capacilities
- Readonly rootFS
- non root user enforced


dduportal (commented):
Helmfile diff looks good!

PR #5580, opened 2024-08-19
fix(publick8s/updates.jio) fine tune configuration: set up correct scan intervals, disable gzip and do not block mirrors when redirecting

dduportal:

This PR fines tunes the mirrors.updates.jio mirrorbits service:
- Disable Gzip (as the ingress already uses gzip or brotli: https://github.com/jenkins-infra/kubernetes-management/blob/e7ce1aa86f72571e29762bb105ada0c6ad65d425/config/public-nginx-ingress__common.yaml#L37-L38) so mirrorbits will use less CPU and will have better latency
- Only scan mirrors once a day since update center already scans every 5 min
- Scan the local repository at the same frequency as the UC script (e.g. 5 min)
- Do NOT disable a mirror if it triggers a redirect
- Set the US East mirror as fallback to handle the 10-15s window every 5 min where local repository is refreshed => this should fix the HTTP/503 errors

### DIFF
--- a/config/updates.jenkins.io.yaml
+++ b/config/updates.jenkins.io.yaml
@@ -100,7 +100,10 @@ mirrorbits:
     checkInterval: 1
     disallowRedirects: false
     disableOnMissingFile: false
-    # No fallbacks
+    fallbacks:
+      - url: https://eastamerica.cloudflare.jenkins.io/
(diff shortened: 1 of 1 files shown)


dduportal (commented):
diff looks good

PR #5584, opened 2024-08-20
fix(infra.ci.jio): init script test for ssh AZ agents

jayfranco999:
As per https://github.com/jenkins-infra/helpdesk/issues/4244#issue-2472582870 :

init script of Linux SSH AZ agents in infra.ci is now in sync with the puppet code. If it tests successful we can proceed with the agent builds https://github.com/jenkins-infra/helpdesk/issues/4124#issuecomment-2295792812

### DIFF
--- a/config/jenkins_infra.ci.jenkins.io.yaml
+++ b/config/jenkins_infra.ci.jenkins.io.yaml
@@ -259,15 +259,20 @@ controller:
                     #!/bin/bash
                     set -eux
                     echo "START CLOUDINIT"
-                    systemctl stop datadog-agent.service
-                    mkdir -p /var/log/datadog /etc/datadog-agent
-                    sed 's/api_key:.*/api_key: ${JENKINS_CI_DATADOG_API_KEY}/' /etc/datadog-agent/datadog.yaml.example > /etc/datadog-agent/datadog.yaml
-                    sed -i 's/# site:.*/site: datadoghq.com/' /etc/datadog-agent/datadog.yaml
-                    chown -R dd-agent:dd-agent /etc/datadog-agent
-                    chmod 640 /etc/datadog-agent/datadog.yaml
-                    chown -R dd-agent:dd-agent /var/log/datadog
-                    chmod 770 /var/log/datadog
-                    systemctl start datadog-agent.service
+                    # Setup Datadog service
+                    (
+                      systemctl stop datadog-agent.service
+                      mkdir -p /var/log/datadog /etc/datadog-agent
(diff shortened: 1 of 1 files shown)


timja (approved):

config/jenkins_infra.ci.jenkins.io.yaml
```diff
@@ -259,15 +259,20 @@ controller:
...
-                    chown -R dd-agent:dd-agent /etc/datadog-agent
-                    chmod 640 /etc/datadog-agent/datadog.yaml
-                    chown -R dd-agent:dd-agent /var/log/datadog
-                    chmod 770 /var/log/datadog
-                    systemctl start datadog-agent.service
+                    # Setup Datadog service
+                    (
+                      systemctl stop datadog-agent.service
+                      mkdir -p /var/log/datadog /etc/datadog-agent
+                      sed 's/api_key:.*/api_key: ^${JENKINS_CI_DATADOG_API_KEY}/' /etc/datadog-agent/datadog.yaml.example > /etc/datadog-agent/datadog.yaml
```
dduportal, line 266:
```suggestion
                      sed 's/api_key:.*/api_key: ${JENKINS_CI_DATADOG_API_KEY}/' /etc/datadog-agent/datadog.yaml.example > /etc/datadog-agent/datadog.yaml
```

The API key is secret value: it's passed to JCasc through a Kubernets Secret, so we want it to be set by JCasc.
As such, no escaping: it MUST be using the interpolation syntax.

Another way to look at it is the following: it was already using the "non escaped" version and was working so there is no reason to change it.

config/jenkins_infra.ci.jenkins.io.yaml
```diff
@@ -259,15 +259,20 @@ controller:
...
+                      sed -i 's/# site:.*/site: datadoghq.com/' /etc/datadog-agent/datadog.yaml
+                      chown dd-agent:dd-agent /etc/datadog-agent/datadog.yaml
+                      chmod 640 /etc/datadog-agent/datadog.yaml
+                      chown dd-agent:dd-agent /var/log/datadog
+                      chmod 770 /var/log/datadog
+                      systemctl daemon-reload
```
dduportal, line 272: Note: I believe this line is the real "root cause" fix and explains why datadog agent was starting flawlessly on ci.jenkins.io but was failing on infra.ci.

dduportal (changes requested):
LGTM! Only one element to fix before merging this (see suggestion).

Good job!

timja (approved):

dduportal (approved):

PR #5648, opened 2024-09-02
feat(publick8s/mirrors.updates.jio) enable `mirrorbits` CLI

dduportal:
This change enables the `mirrorbits` CLI and exposes the server gRPC protocol through a Kubernetes Service.

This service of `LoadBalancer` kind associated to an Azure Private Link service: it allows access through private networks without requiring any network peering of any kind.

Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2325103688

Requires:

- Helm chart support for the `mirrorbits` CLI: https://github.com/jenkins-infra/helm-charts/pull/1291
- Permission to be fixed to allow `publick8s` to manage the Azure PLS: https://github.com/jenkins-infra/azure/pull/819

PR #5653, opened 2024-09-03
feat(publick8s/updates.jenkins.io) switch PLS to the data tier subnet

dduportal:
Follow up of #5648 : the PLS is not reachable as the AKS subnet has a custom routing table which create mayhem.

We also should be safer by having this resource with a con figuration as close as possible as another internal LB working (private Nginx ingress).

Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2325103688


smerle33 (approved):
🚀

PR #5665, opened 2024-09-05
feat(ALL/nginx ingresses): add namespace and ingress name to the access logs

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2330934069

The ingress (access) logs collected in datadog does not provide an easy way to discriminate between services.

This PR adds the namespace and name of the ingress for which a given access log line is meant to.

It follows the Nginx Ingress controller documentation which provides these 2 fields in addition to the usual Nginx fields: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/log-format/

Tested on a manual apply (and then roll backed) in pair with @smerle33 on the private Nginx of the private cluster, to ensure it does not fail the admission + the access logs format is changed with what we want

### DIFF
--- a/config/private-nginx-ingress__common.yaml
+++ b/config/private-nginx-ingress__common.yaml
@@ -23,7 +23,7 @@ defaultBackend:
       emptyDir: {}
 controller:
   config:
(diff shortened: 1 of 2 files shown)


smerle33 (approved):
🚀

Issue #5663, opened 2024-09-05
[updatecli] Helm Chart-related manifests: switch targets from (tedious) file regexs to (native) YAML queries


dduportal:
### What feature do you want to see added?

This repository describes Kubernetes cluster expected applications using helmfile releases.

The YAML file used to describe these "helm releases" uses a list of object as you can see in https://github.com/jenkins-infra/kubernetes-management/blob/main/clusters/publick8s.yaml

We are tracking the helm chart versions with a set of updatecli manifests in https://github.com/jenkins-infra/kubernetes-management/tree/main/updatecli/updatecli.d/charts which role is to detects a new chart versions and open a PR with new versions in the (aforementioned) "helm releases YAML" file.

Problem: due to the inherent fragility of using numeric index (can be reordered, breaking the updatecli manifests), we switched to updatecli [targets](https://www.updatecli.io/docs/core/target/) of kind ["file"](https://www.updatecli.io/docs/plugins/resource/file/) which uses a regex to match patterns and proceed to replacement such as https://github.com/jenkins-infra/kubernetes-management/blob/f38be834e33b1790abd3b7f8cf4a0c4e1a48af92/updatecli/updatecli.d/charts/accountapp.yaml#L30-L31.

=> this is hard and tedious to maintain, and prone to error.

Proposed solution: updatecli did now support a yaml path engine in its [YAML]() target kind. It allows using YAML path queries to perform selection of the list object based on attribute filters which is way better.

Working example: https://github.com/jenkins-infra/kubernetes-management/blob/main/updatecli/updatecli.d/charts/artifact-caching-proxy.yaml

=> we want to update all manifest tracking Helm Charts versions to use YAML path targets instead of file targets, following the working example

dduportal:
Closed by https://github.com/jenkins-infra/kubernetes-management/pull/5666

PR #5666, opened 2024-09-05
feat(updatecli): Switch Helm Chart related manifest targets to native(YAML) queries

jayfranco999:
As per https://github.com/jenkins-infra/kubernetes-management/issues/5663#issue-2506884201:

We updated all manifest tracking Helm Charts versions to use YAML path targets instead of file targets.

It allows using YAML path queries to perform selection of the list object based on attribute filters which was the proposed solution.

### DIFF
--- a/updatecli/updatecli.d/charts/accountapp.yaml
+++ b/updatecli/updatecli.d/charts/accountapp.yaml
@@ -23,7 +23,7 @@ sources:
 targets:
   updateChartVersion:
     name: "Update the chart version for accountapp"
-    kind: file
+    kind: yaml
     scmid: default
     spec:
       file: clusters/publick8s.yaml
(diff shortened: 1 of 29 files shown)


jayfranco999:
Files not updated:

updatecli/updatecli.d/charts/httpd.yaml
updatecli/updatecli.d/charts/jenkins-agents-infra.ci.jenkins.io.yaml
updatecli/updatecli.d/charts/jenkins-infra-tools-maven.yaml

updatecli/updatecli.d/charts/accountapp.yaml
```diff
@@ -29,7 +29,6 @@ targets:
       file: clusters/publick8s.yaml
       engine: yamlpath
       key: $.releases[?(@.name == 'accountapp')].version
-    scmid: default
```
dduportal: Sounds like a mistake: you should not remove this instructions. Otherwise the target would never be applied to the git repository

dduportal: 👍

dduportal (changes requested):
As see together: there a changes missing (`kind`)  otherwise it won't work as expected

jayfranco999:
`kind: yaml` has been fixed for all the Helm Chart related manifests.

PR #5680, opened 2024-09-12
fix(publick8s/updates.jenkins.io) force redirection on mirrorbits even if accept header is set to 'application/json'

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2345352162

Requires https://github.com/jenkins-infra/helm-charts/pull/1327 feature


dduportal (commented):
diff looks good

PR #5708, opened 2024-09-18
feat(publick8s/updates.jenkins.io) add an HTTP-only mirrorbits instance

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2352944136

Need the new (Redis) secret: https://github.com/jenkins-infra/charts-secrets/commit/66340f36ab34a678e4430a8aee36d203e3255075

This PR installs, in updates-jenkins-io, a new mirrorbits instance which uses the same data as the existing one but has different ingresses.

It relies on 2 new mirrorbits chart features:
- Ability to reuse the same PVCs: https://github.com/jenkins-infra/helm-charts/pull/1339
- Ability to specify additional ingresses: https://github.com/jenkins-infra/helm-charts/pull/1344

The support of ingress traffic split based on the protocol used (HTTP or HTTPS) utilizes the following techniques:

- The existing ingress for `mirrors.updates.jenkins.io`, which handles both HTTP and HTTPS (and will continue):
  - Does not force SSL redirection anymore (obviously)
  - Has a new configuration-snippet which rewrite internally (`rewrite <...> <...> last`) requests to `/internal_http` when the request scheme is `http` AND is not already `/internal_http` (to avoid infinite redirections)
- The new ingress rule:
  - Is set to only handle HTTP as it does not have any `tls` block (in real life, the Nginx Ingress controller will merge it with the previous one)
  - Only capture requests from `/internal_http` path and rewrite them before sending them to the new HTTP only SVC/Pods mirrorbits
  - Note we require a secondary HTTP-only ingress to manage the rewrite from `/` to `/index.html` in HTTP mode.

----

Notes:

- We use the `$uri` Nginx variable instead of `$request_uri` because the latter one is not updated in internal redirects
- We cannot use if with AND / OR operator. Instead we use a technique with a variable used to evaluate the condition.
- The mirrorbits database Redis is a new one. We've already enabled it during manual test.

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -227,6 +227,14 @@ releases:
       - "../config/updates.jenkins.io.yaml"
     secrets:
       - "../secrets/config/updates.jenkins.io/secrets.yaml"
+  - name: updates-jenkins-io-unsecured
+    namespace: updates-jenkins-io
+    chart: jenkins-infra/mirrorbits
+    version: 6.0.7
+    values:
+      - "../config/updates.jenkins.io-unsecured.yaml"
+    secrets:
+      - "../secrets/config/updates.jenkins.io/secrets-unsecured.yaml"
   - name: updates-jenkins-io-httpd
     namespace: updates-jenkins-io
     chart: jenkins-infra/httpd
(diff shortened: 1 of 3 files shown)


smerle33 (approved):
LGTM 🚀

PR #5720, opened 2024-09-20
feat(stats.jenkins.io): override the locations to avoid 200 for 404 errors

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4132#issuecomment-2352333521 and https://github.com/jenkins-infra/helpdesk/issues/4291

lets defined manual locations to avoid 200 errorcode for 404

### DIFF
--- a/config/stats.jenkins.io.yaml
+++ b/config/stats.jenkins.io.yaml
@@ -57,32 +57,24 @@ affinity:
 
 nginx:
   overrideLocations: |
-    location /statistics {
+    location /pluginversions {
         root   /usr/share/nginx/html;
         index  index.html index.htm;
         autoindex on;
-        try_files $uri /index.html;
     }
-    location /plugin-trends {
+    location /plugin-installation-trend {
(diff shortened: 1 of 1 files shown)


timja (commented):
This doesn't seem that scaleable, any new routes will have to be added here and it will be easily missed.

Would it be possible to do this in the inverse so its only not enabled for the actual static file?

smerle33:
> This doesn't seem that scaleable, any new routes will have to be added here and it will be easily missed.
> 
> Would it be possible to do this in the inverse so its only not enabled for the actual static file?

very good point, let me reverse the logic

smerle33:
as per https://github.com/jenkins-infra/stats.jenkins.io/pull/168/files I reverse the logic to handle 404 only on those path:
```
https://stats.jenkins.io/pluginversions
https://stats.jenkins.io/plugin-installation-trend
https://stats.jenkins.io/jenkins-stats

PR #5736, opened 2024-09-24
feat(publick8s/azure.updates.jio) switch https (securedd) Apache to the new fileshare

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2364054995

This PR introduces the following changes to the `publick8s`'s Apache release for `azure.updates.jenkins.io` service:

- Rename release and config for clarity and symmetry with the "unsecured" Apache installation of the same service
- Bump the chart version to 1.3.0 as it's not tracked (ping @jayfranco999 for info) as we need latest chart release (subpath support) here
- Switch the File share to the new one (proper name + subpath `./secured`)

Note: The subpath and PVC name can now be retrieved from https://reports.jenkins.io/jenkins-infra-data-reports/azure.json by `updatecli` (ping @jayfranco999 for info)


dduportal:
This PR has a diff which looks good *BUT* is blocked by:

- Need https://github.com/jenkins-infra/helm-charts/issues/1356 to be solved and the HTTPD chart to be updated here to > 1.3.0
- Need the current release to be manually uninstalled otherwise it will fail when applying as the objects already exist

dduportal (commented):
- version 1.3.1 is used
- Former released uninstalled with `helm uninstall updates-jenkins-io-httpd -n updates-jenkins-io`

LGTM let's roll

PR #5903, opened 2024-11-19
fix(publick8s/updates.jio) disable ingrerss rewrite logs as it is resources hungry

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/2649


dduportal (commented):
LGTM, diff is good

PR #6073, opened 2025-01-03
feat(cijenkinsio-agents-2) initial installation of ACP

dduportal:


dduportal (commented):
Let's roll: tested once (and then NS was deleted) with success

PR #6084, opened 2025-01-08
fix(publick8s) ensure cert-manager tracks the SSL certificate for mirrors.updates.jenkins.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4481

This PR adds a missing annotation which allows Cert Manager to track the certificate and renew it


dduportal (commented):
Helmfile Diff looks good!

PR #6359, opened 2025-03-25
fix(ci.jenkins.io-agents-2) disable blob caching as a hack to avoid corrupted zero-sized blobs

dduportal:
… - https://github.com/distribution/distribution/issues/2367\#issuecomment-1874449361

### DIFF
--- a/config/hub-mirror_cijioagents2.yaml
+++ b/config/hub-mirror_cijioagents2.yaml
@@ -59,3 +59,8 @@ initContainers:
     volumeMounts:
       - mountPath: /var/lib/registry
         name: data
+
+extraEnvVars:
+  # Ref. https://github.com/distribution/distribution/issues/2367#issuecomment-1874449361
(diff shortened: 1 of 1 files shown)


timja:
@dduportal adding a backslash breaks the link, if you're trying to avoid backlinks use backticks, otherwise something is going wrong in pasting, (I fixed this earlier in a PR you did about nginx ingress-controller as well)

dduportal:
> @dduportal adding a backslash breaks the link, if you're trying to avoid backlinks use backticks, otherwise something is going wrong in pasting, (I fixed this earlier in a PR you did about nginx ingress-controller as well)

Thanks! Looks like the macOS terminal has a new behavior in which it escapes the links when copy and pasted (into a commit message for instance). I'll be carefull.

dduportal:
In fact is is ZSH: `DISABLE_MAGIC_FUNCTIONS=true` in my `.zshrc` fixed the problem.

PR #6401, opened 2025-04-02
feat(cijenkinsio-agents-1) add cluster with datadog only

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4619

This PR adds back the management of the cluster `cijenkinsio-agents-1` in Azure:

- It starts with a revert of https://github.com/jenkins-infra/kubernetes-management/pull/6282 (had to reverte the 2 commits https://github.com/jenkins-infra/kubernetes-management/commit/dda4c54c9e372fc1a8738d01d54922bc1de0b191 and https://github.com/jenkins-infra/kubernetes-management/commit/f84ee11cafb3eb64cf724e433eb3b31d06f48d30)
- It adds back the cluster in the pipeline
- It only installs datadog release (for now: let's go step by step) with:
  - Fixed cluster name/file (to be coherent with the EKS cijenkinsio-agents-2 cluster)
  - Updated datadog chart version + tracking it in updatecli
  - Deploy on `amd64` (ref. https://github.com/jenkins-infra/azure/pull/977)

PR #6402, opened 2025-04-02
fix(cijenkinsio-agents-2) ensure the proper CPU architecture is used …

dduportal:
…- fixup of #6401


dduportal:
Helmfile diff on cijenkinsio-agents-2 looks good!

PR #6403, opened 2025-04-02
feat(cijenkinsio-agents-1) install ACP with 1 replica and no CPU limits

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4619

The ACP settings have been extracted from the former Azure cluster, but adapted to what we learnt when running in AWS (ref. https://github.com/jenkins-infra/helpdesk/issues/4545):

- Only 1 replica
- No CPu limit to avoid throttling requests (it is safe to do this on non CI agents workloads when we have static workloads)


dduportal:
Helmfile diff looks good for cijenkinsio-agents-1, and no changes for the other clusters

PR #6469, opened 2025-04-22
fix(publick8s/weekly.ci) remove deprecated parameter for the custom-header plugin

dduportal:
Ref. https://github.com/jenkinsci/customizable-header-plugin/pull/194/files#r2054468199


smerle33 (approved):
🚀

PR #6519, opened 2025-04-28
feat(privatek8s-sponsorship) add public-nginx-ingress release

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4250#issuecomment-2824642025

Tested with success manually, then uninstalled (and removed NS) prior to the PR. The public LB was created with the provided public IPv4 with no errors.


dduportal:
Helmfile diff looks good

PR #6520, opened 2025-04-28
feat(privatek8s-sponsorship) add core-package PV/PVCs

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4250#issuecomment-2824642025

This PR adds the 2 PVCs (and associated resources: PVs and secret) used by release.ci agents.

Note: the goal is to create the resources with the same settings as in the current `privatek8s` before importing them in Terraform for static management (so it would be easier to check the settings and eventual drift in Terraform by importing them)


dduportal (commented):
Already created manually: let's roll

PR #6682, opened 2025-06-18
feat(publick8s/reports.jio) switch to the "common" nginx-webserver chart to allow custom pod annotation for datadog logs collection

dduportal:

Note: this PR changes the chart so it will recreate ressources. The goal is to get rid of the `jenkin-infra/reports` chart which is only duplicating the `nginx-webserver` but with less features.
It's easier to provide the custom annotation (even if we could contribute the `podAnnotation` helm value in `report`)

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -165,8 +165,8 @@ releases:
       - "../secrets/config/uplink/secrets.yaml"
   - name: reports
     namespace: reports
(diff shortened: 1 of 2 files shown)


dduportal:
Almost there: this PR required an hotfix (https://github.com/jenkins-infra/kubernetes-management/commit/8f2f29813b35550c0c0036385e043ab2b8a4e2ea) to complete its deployment

PR #6690, opened 2025-06-19
cleanup(publick8s) remove jenkinsistheway.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4709

The goal is to cleanup the cert-manager logs after https://github.com/jenkins-infra/kubernetes-management/pull/6644 so better to remove this non functional service


dduportal (commented):
Weird: the helmfile diff shows removal of 2 ClusterIssuers instead of 1 as expected. Do not merge

dduportal:
> Weird: the helmfile diff shows removal of 2 ClusterIssuers instead of 1 as expected. Do not merge

Should be fixed by https://github.com/jenkins-infra/kubernetes-management/pull/6690/commits/cd0e30f08853bb34df91b54af613e6c62f98a468 (which requires https://github.com/jenkins-infra/charts-secrets/commit/29d8da6051fa9705e79314905628445ba1a0b15a)

ACME chart seems to not have deeply merged values. So I'm reverting status back to "1 only value file"

PR #6693, opened 2025-06-19
feat(publick8s/javadoc.jio) switch to the "common" nginx-webserver chart to allow custom pod annotation for datadog logs collection

dduportal:

Same idea as #6682 but for `javadoc.jenkins.io`

- Note: Did change the volume secret name in the config (this time) to avoid a hotfix such as 8f2f2981
- chore: moved updatecli chart version tracking of javadoc to the "nginx-webserver" manifest (no need for subsequent PR)


dduportal:
Helmfile diff looks good!

PR #6709, opened 2025-06-19
fix(publick8s/updates.jenkins.io) use the proper datadog logs parser for httpd

dduportal:

Ref. https://docs.datadoghq.com/integrations/apache/?tab=host


dduportal:
helmfile diff is good 👍

PR #6872, opened 2025-07-30
cleanup: remove all remnants of privatek8s

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4690


dduportal (commented):
- Updatecli checks updated (many manifests improved and cleaned up)
- Only 2 helmfile diffs: LDAP (removing unused IPs from privatek8s-sponsorship) and removing credential from infra.ci

PR #6878, opened 2025-08-01
feat(cijenkinsioagents2/maven-cacher) fix cache generation by adding script fix and increase resources

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4752


dduportal (commented):
Helmfile diff looks good

PR #6945, opened 2025-08-13
feat(publick8s/updates.jenkins.io) use the same NFS volume in content for both mirrorbits repository and geoipdata

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4767#issuecomment-3180139293

Follows up #6942 and #6943 

Requires https://github.com/jenkins-infra/helm-charts/pull/1761

=> this PR changes the `geoipdata` mount to use the sub directory `updates.jenkins.io/geoipdata` from the common NFS volume. This directory has already been populated, but is NOT kept up to date yet (update once a week usually).

### DIFF
--- a/clusters/publick8s.yaml
+++ b/clusters/publick8s.yaml
@@ -218,7 +218,7 @@ releases:
   - name: updates-jenkins-io-content
     namespace: updates-jenkins-io
     chart: jenkins-infra/mirrorbits
-    version: 5.8.2
+    version: 5.9.0
     values:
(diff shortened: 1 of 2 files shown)


dduportal:
- Note: updatecli fails with the same error as main branch. Merging this PR will fix the error below:

```
something went wrong in target "updateGeoipdataPvcSubdir" : "updating yaml file: couldn't find key \"$.geoipdata.subDir\" from file \"config/updates.jenkins.io-content.yaml\""
```

as the key is created

- Helmfile diff looks good

PR #6948, opened 2025-08-13
feat(publick8s) create staging.get.jenkins.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4767#issuecomment-3180139293

- Both mirrorbits and httpd but without the "mirrorbits-parent" chart
  - Distinct charts: easier to manage (shorter feedback loop, less code)

- mirrorbits v0.6.0

- New NFS data store
  - ReadonlyMany
  - Proper Storage class (no typo)
  - Better performances
  - Timestamp kept


dduportal:
Self merging as already applied manually

PR #6958, opened 2025-08-15
feat(publick8s/get.jenkins.io) switch persistent volume for mirrorbit…

dduportal:
…s and httpd to the new data-storage common NFS

Ref. https://github.com/jenkins-infra/helpdesk/issues/4767


dduportal:
Helmfile diff shows expected changes from the code:

- mirrorbits containers have their "data" volume changed to the new NFS with a subpath (same as staging.get.jio)
- httpd containers have their "data" volume changed to the new NFS with a subpath (same as staging.get.jio)
- The old PV and PVC are removed (but the backend storage is retained)

PR #6959, opened 2025-08-15
feat(publick8s/get.jenkins.io) switch persistent volume for geoipupdate to the new data-storage common NFS

dduportal:
Follow up #6958 . It's the same but only for the geoipdata persistent volume.

Ref. https://github.com/jenkins-infra/helpdesk/issues/4767

- The Staging service (staging.get.jenkins.io) shows that we have a working data storage with the provided attributes: https://github.com/jenkins-infra/helpdesk/issues/4764#issuecomment-3187400908 (geoip databases included)

### DIFF
--- a/config/get-jenkins-io.yaml
+++ b/config/get-jenkins-io.yaml
@@ -119,20 +119,8 @@ mirrorbits:
         countryCode: DE
         continentCode: EU
   geoipdata:
-    persistentData:
-      enabled: true
-      capacity: 1Gi
-      storageClassName: statically-provisionned
-      csi:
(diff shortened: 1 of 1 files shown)


dduportal:
- The helmfile diff shows the exact expected changes from the code:
  - The mirrorbits pods now mount the new subpath from the current `data` volume (which is already on the NFS storage since #6958) and removed volume `geoipupdate
  - PV and PVC for geoipupdate are reclaimed
- Note: updatecli checks fails. A subsequent PR will fix it once for all though.

PR #6960, opened 2025-08-15
cleanup(publick8s/get.jenkins.io) scale down to 1 replicas for each component as a preliminary for migration to split helm releases

dduportal:
Preliminary to https://github.com/jenkins-infra/kubernetes-management/pull/6960:

- Removes the PDB (easier to start new pods)
- Make the incoming "sync" part faster to run


dduportal:
Helmfile diff is good

PR #7037, opened 2025-09-04
fix(privatek8s/infra.ci.jenkins.io) ensure pipeline library uses git SCM withtou credential with PR support and correct GHap top level credential scope

dduportal:
Revert #7033


dduportal (commented):
Already applied manually

PR #7086, opened 2025-09-24
fix(publick8s) use coherent naming/NS for wiki.jenkins.io

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4617

- Already applied manually which created the namespace `wiki-jenkins-io`
- Removed the former `wiki` namespace


dduportal:
Helmfile diff looks good (no changes anywhere)

PR #7093, opened 2025-09-25
fix(publick8s) expose private ingress to the cluster subnet

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4617

We want to expose the private accesses (Azure PLS for updates.jenkins.io, and the private ingress) in the same subnet as the cluster itself


dduportal:
Helmfile diff is good

PR #7616, opened 2026-04-20
feat(publick8s/updates.jenkins.io) allow private endpoint access to mirrorbits CLI Private Link Service from sponsored subscription

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/5070#issuecomment-4268303462

This PR updates the Private Link Service used to expose the `mirrorbits` CLI for updates.jenkins.io in `publick8s` to trusted.ci.jenkins.io by allowing access from the new sponsored subscription.

It should fix https://github.com/jenkins-infra/azure/pull/1407#issuecomment-4278362223 (Private Endpoint error creation in the new subscription).

Note: https://learn.microsoft.com/en-us/azure/aks/internal-lb?tabs=set-service-annotations for reference


dduportal:
Helmfile diff looks good! Merging

PR #7630, opened 2026-04-27
feat(privatek8s/infra.ci.jenkins.io) restrict Gatsby job to only production deployment

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5062#issuecomment-4327443012

This PR updates the configuration of the `gatsby-plugin-jenkins-layout` Multibranch job on infra.ci.jenkins.io with the following changes:

- Disable pull request builds (handled by ci.jenkins.io as no Netlify preview is needed)
- Remove `allowUntrustedChanges` attribute (implicitly set to `false` as there are no PRs anymore)
- Explicitly specify `disableTagDiscovery` to `true` as it seems to have tag based builds
- Move out of the "Website Jobs" folder to "Website production Jobs" folder to ensure credentials are a bit more isolated (following #4282)

In order to keep the jobs history, i'll backup the current job on disk and I'll restore the main's branch builds along with the tags builds *before* merging this. Then we'll let JCasC do its magic, and we'll reload from disk.

### DIFF
--- a/config/jenkins-jobs_infra.ci.jenkins.io.yaml
+++ b/config/jenkins-jobs_infra.ci.jenkins.io.yaml
@@ -684,21 +684,6 @@ jobsDefinition:
             appId: "${GITHUB_APP_JENKINS_IO_COMPONENTS_ID}"
             owner: "jenkins-infra"
             privateKey: "${GITHUB_APP_JENKINS_IO_COMPONENTS_PRIVATE_KEY}"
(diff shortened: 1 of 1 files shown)


dduportal:
- Backup taken (`.backups/20260427-gatsby-plugin-jenkins-layout.tgz`) in data disk and verified (`tar tzf ...`)
- Helmfile diff looks good.

PR #7640, opened 2026-04-28
fix(publick8s/datadog/custom-checks): parse Unix epoch timestamp in build report staleness check

jayfranco999:
Ref:

- https://github.com/jenkins-infra/helpdesk/issues/2843

The build report staleness check (`BuildReportStaleness`) fails to parse  `report_timestamp` from `status.json`, causing `age_in_hours`, `threshold_in_hours`, and `stale` metrics to never be emitted.

The report generator shell script was updated in jenkins-infra/pipeline-library#956 to use Unix epoch format (`date -u +%s`) instead of ISO 8601, but the agent  check was never updated to match, it still calls `datetime.fromisoformat()`.

Issue diagnosed on datadog-agent logs:

image

### DIFF
--- a/config/datadog_confd_checksd.yaml
+++ b/config/datadog_confd_checksd.yaml
@@ -195,7 +195,7 @@ datadog:
 
               # Age (staleness)
               try:
-                  ts = datetime.fromisoformat(data['report_timestamp'].replace('Z', '+00:00'))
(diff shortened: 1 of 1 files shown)


dduportal (approved):
🚀

PR #7696, opened 2026-05-12
feat(privatek8s/infra.ci.jenkins.io) split the plugins.jenkins.io (plugin-site) project into 2 distinct jobs (PR previews and deployment)

dduportal:
Same as https://github.com/jenkins-infra/kubernetes-management/pull/7695 for plugins.jenkins.io (plugin-site):

Let's split the plugins.jenkins.io project, in infra.ci.jenkins.io jobs, into 2 distinct jobs:

- 1 for PR previews
  - Will be triggered by external contributors (It reverts the safety setup put in place in https://github.com/jenkins-infra/helpdesk/issues/4282 as no more credentials in this job except the Netlify deploy token) 
- 1 for production deployment

PR #7695, opened 2026-05-12
feat(privatek8s/infra.ci.jenkins.io) split the stories.jenkins.io project into 2 distinct jobs (PR previews and deployment)

dduportal:
Following #7694, let's split the stories.jenkins.io project, in infra.ci.jenkins.io jobs, into 2 distinct jobs:

- 1 for PR previews (only Netlify credential).
  - Will be triggered by external contributors (It reverts the safety setup put in place in https://github.com/jenkins-infra/helpdesk/issues/4282 as no more credentials in this job except the Netlify deploy token) 
- 1 for production deployment
  - Note: production is currently in Netlify but [it should move to our own hosting](https://github.com/jenkins-infra/helpdesk/issues/4958) one of these days

PR #7694, opened 2026-05-12
feat(privatek8s/infra.ci.jenkins.io) adapt configuration of the 2 stats.jenkins.io website jobs

dduportal:
Following #7633 and #7690, this PR updates the configuration of the stats.jenkins.io website jobs in infra.ci.jenkins.io:

- The "PR" previews job:
  - Won't build branches anymore. Only PRs as expected.
  - Will be triggered by external contributors (It reverts the safety setup put in place in https://github.com/jenkins-infra/helpdesk/issues/4282 as no more credentials in this job except the Netlify deploy token)
- The "production" deployment job won't build PRs anymore

PR #7690, opened 2026-05-12
feat(privatek8s/infra.ci.jenkins.io/website-jobs) allow untrusted jobs for external contributors on contributor-spotlight website preview

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4985

Following https://github.com/jenkins-infra/kubernetes-management/pull/7633, we can re-enable the untrusted as we don't have credential anymore in the PR jobs except Netlify's token.

It reverts the safety checks put in place in https://github.com/jenkins-infra/helpdesk/issues/4282 but only for this job

PR #7699, opened 2026-05-12
fix(privatek8s/infra.ci.jenkins.io) various naming, setup and cleanup on the website jobs

dduportal:
Following #7690, #7694, #7695, #7696, #7697 and #7698, this PR applies a few minor fixes on the website jobs in infra.ci.jenkins.io:

- Lexicographic Ordering
- Name/Description convention
- Remove explicit default attributes to have the same convention as other jobs (and a bit less verbose)
- Fix a missing credential for jenkins-io-components


dduportal (commented):
Helmfile diff looks good 👍

PR #7725, opened 2026-05-22
feat: add `privatek8s-sponsored` with `datadog` release

lemeurherve:
This change adds the new cluster, and starts putting its config files in their own folder, with common values staying at the root of ./config folder.

Refs:
- https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4381409952
- https://github.com/jenkins-infra/charts-secrets/commit/6c5fcc4658fcdd03c8e328df473ad93c3afac5fb (for datadog secret)

PR #7732, opened 2026-05-22
chore(publick8s): regroup config files in `./config/publick8s` folder

lemeurherve:
Similar organisation as:
- #7725 

A follow-up PR will adapt updatecli manifests accordingly.


lemeurherve:
No diff as expected 🙂

dduportal (approved):

PR #7733, opened 2026-05-22
chore(privatek8s-sponsored): use helmfile state values

lemeurherve:
Could be quite useful when switching clusters.

Ref:
- https://helmfile.readthedocs.io/en/latest/writing-helmfile/#helmfile-values-vs-helm-values


config/datadog.yaml.gotmpl
```diff
@@ -16,18 +21,11 @@ datadog:
     containerCollectAll: true
   networkMonitoring:
     enabled: true
-  {{- if .Values }}
-  {{- if hasKey .Values.datadog "kubelet" }}
```
lemeurherve, line 20: No mention of any `kubelet` anywhere except in this template. Removed as it created some lint errors.

config/privatek8s-sponsored/datadog.yaml
```diff
@@ -1,13 +1,3 @@
-providers:
-  aks:
-    enabled: true
-datadog:
-  clusterName: 'privatek8s-sponsored'
-  env:
-    - name: DD_HOSTNAME
-      valueFrom:
-        fieldRef:
-          fieldPath: spec.nodeName
```
lemeurherve, line 10: Identical section in all cluster datadog values, moved to common datadog values.

dduportal:
Closing as this was a PoC. We'll restart the work if needed, but the conflicts + scope definition should be defined more precisely.

Thanks for the pointers!

PR #7749, opened 2026-05-26
feat(privatek8s-sponsored/release.ci.jenkins.io) add empty controller (no jobs, initial data volume fsgroup AKS CSI setup and no startup probes)

dduportal:

Requires #7742 

This PR initializes the release.ci.jenkins.io as an empty controller to ensure it can start in the new cluster, prior to verifying its "acceptance" test jobs passes.

As such it has the following properties which will be changed before production but after validations:

- Data disk is empty and new. As such it requires the AKS CSI driver to set up the permissions: fsgroup and associated configs are enabled (one time thing)
  - On production migration, we'll rely on the permissions from the `privatek8s` controller data disk
- No startup probes (for now)
- Empty jobs section (we don't want concurrent jobs execution with production)

----

(edited) Will need subsequent PR (one or many) before production to:
- Disable the fsgroup AKS CSI stuff
- Enable probes
- Set up the "acceptance test" job as code (JobDSL)

### DIFF
--- a/config/privatek8s-sponsored/release.ci.jenkins.io.yaml
+++ b/config/privatek8s-sponsored/release.ci.jenkins.io.yaml
@@ -1,14 +1,17 @@
 serviceAccount:
+  # Managed by Terraform to allow enabling Federated Identities
   create: false
   # TODO: track with updatecli from https://reports.jenkins.io/jenkins-infra-data-reports/azure.json
   name: release-ci-jenkins-io-controller
 serviceAccountAgent:
+  # Managed by Terraform to allow enabling Federated Identities
   create: false
 rbac:
   create: true
   readSecrets: true
 persistence:
   enabled: true
+  # TODO: track with updatecli from https://reports.jenkins.io/jenkins-infra-data-reports/azure.json
   existingClaim: release-ci-jenkins-io-data
 agent:
   componentName: "agent"
@@ -23,6 +26,7 @@ networkPolicy:
   internalAgents:
     allowed: true
     namespaceLabels:
+      # TODO: track with updatecli from https://reports.jenkins.io/jenkins-infra-data-reports/azure.json
       name: "release-ci-jenkins-io"
 controller:
   image:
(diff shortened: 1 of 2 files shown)


config/privatek8s-sponsored/release.ci.jenkins.io.yaml
```diff
@@ -0,0 +1,506 @@
...
+                webSocket: true
+                name: "kubernetes"
+                namespace: "release-ci-jenkins-io-agents"
+                podRetention: "Never"
+                podLabels:
+                  - key: "jenkins/release-ci-jenkins-io-agent"
+                    value: "true"
+                  - key: "azure.workload.identity/use"
+                    value: "true"
+      # jobs-settings: |
```
lemeurherve, line 196: AFAICT, there is no config as code for https://release.ci.jenkins.io/job/infra-agents-health

Should we define release.ci.jenkins.io jobs as code in their own config file like what's done for infra.ci.jenkins.io?

dduportal: > AFAICT, there is no config as code for https://release.ci.jenkins.io/job/infra-agents-health

Yep, came to the same discovery when preparing this PR earlier today.

> Should we define release.ci.jenkins.io jobs as code in their own config file like what's done for infra.ci.jenkins.io?

Absolutely: but since we' ll need a second PR to disable the AKS CSI `fsGroup` stuff and to enable the probes, I'll add the health check on this subsequent PR, so we can already check that the empty controller starts (annd azure resource permissions are ok, etc.)

lemeurherve (approved):

dduportal:
@lemeurherve I've reworked the commit history as per your request to make it easier to review

lemeurherve (approved):

dduportal:
Deployed with success!

I can access it by adding the following line in my `/etc/hosts`:

```
# private.privatek8s-sponsored.jenkins.io
10.6.0.254      release.ci.jenkins.io
```

and then I get the following in my web browser:

Capture d’écran 2026-05-26 à 14 39 20

PR #7757, opened 2026-05-27
feat(privatek8s-sponsored/release.ci.jenkins.io) add `infra-agents-health` job

dduportal:

This PR defines the `infra-agents-health` as code for release.ci.jenkins.io controller in the new `privatek8s-sponsored` cluster.

It introduces a few changes compared to its "manually-managed" original twin from `privatek8s`:

- Type is a Multibranch job with restricted branch name, no PR, no tags and no github checks (instead of single pipeline job)
  - Note: the credential is set up to an empty value which means it might be rate limited 
- It uses the source branch of https://github.com/jenkins-infra/release/pull/908 as there are changed parameters (node pool named) compared to the original. We'll merge and update once migration is performed

### DIFF
--- a/config/privatek8s-sponsored/release.ci.jenkins.io.yaml
+++ b/config/privatek8s-sponsored/release.ci.jenkins.io.yaml
@@ -193,8 +193,78 @@ controller:
                     value: "true"
                   - key: "azure.workload.identity/use"
                     value: "true"
-      # jobs-settings: |
-      #   jobs:
+      jobs-settings: |
+        jobs:
+          - script: >
+              multibranchPipelineJob('infra-agents-health') {
+                description('ddu-6')
+
+                triggers {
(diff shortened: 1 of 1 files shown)


config/privatek8s-sponsored/release.ci.jenkins.io.yaml
```diff
@@ -193,8 +193,78 @@ controller:
...
                   - key: "azure.workload.identity/use"
                     value: "true"
-      # jobs-settings: |
-      #   jobs:
+      jobs-settings: |
+        jobs:
+          - script: >
+              multibranchPipelineJob('infra-agents-health') {
+                description('ddu-6')
+
```
dduportal:
```suggestion
```

dduportal:
Tested manually with success: let's merge!

PR #7758, opened 2026-05-27
 feat(privatek8s-sponsored/infra.ci) init. empty controller (initial data volume fsgroup AKS CSI setup and no startup probes)

dduportal:

Same as #7749 but for infra.ci.jenkins.io: This PR initializes infra.ci.jenkins.io as an empty controller to ensure it can start in the new cluster, prior to verifying its "acceptance" test jobs passes.

As such it has the following properties which will be changed before production but after validations:

- Data disk is empty and new. As such it requires the AKS CSI driver to set up the permissions: fsgroup and associated configs are enabled (one time thing)
  - On production migration, we'll rely on the permissions from the `privatek8s` controller data disk
- No startup probes (for now)
- The `infra-ci-jenkins-io-jobs` release is also updated to get the latest setup of the acceptance test (from #7754 and #7755).

----

Note: Will need subsequent PR (one or many) before production to:
- Disable the fsgroup AKS CSI stuff
- Enable probes

### DIFF
--- a/clusters/privatek8s-sponsored.yaml
+++ b/clusters/privatek8s-sponsored.yaml
@@ -70,21 +70,21 @@ releases:
     version: 3.2.0
     values:
       - ../config/privatek8s-sponsored/infra-ci-jenkins-io-jobs.yaml
-  # - name: infra-ci-jenkins-io
-  #   namespace: infra-ci-jenkins-io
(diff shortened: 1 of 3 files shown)


dduportal:
Helmfile diff looks good!

PR #7759, opened 2026-05-27
fix(privatek8s-sponsored/release.ci) configuration fixes: use git for health job and send advisory to team mailing list

dduportal:
Fixup of #7757 to avoid GitHub API rate limit

Also a minor config along the way


dduportal (commented):
Helmfile diff looks good!

PR #7772, opened 2026-05-27
cleanup: remove privatek8s in favor of privatek8s-sponsored

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4554160116

- Removes all remnants of `privatek8s` configurations
- Fix `updatecli` manifests
- Removes the old privatek8s NAT from LDAP
- Removes the former kubeconfig credential (need to remove from SOPS though)


dduportal:
Helmfile diff looks good

PR #7778, opened 2026-05-29
chore(publick8s/datadog): extract scripts and data files from datadog_confd_checksd.yaml

lemeurherve:
This change extracts python scripts and data YAML to a folder instead of inlining them in a YAML, for better readability and maintainance.

### Testing done

```bash
$ helmfile template -f ./clusters/publick8s.yaml -l name=datadog > datadog_confd_checksd-AFTER.yaml
$ git checkout main
$ helmfile template -f ./clusters/publick8s.yaml -l name=datadog > datadog_confd_checksd-BEFORE.yaml
$ diff -u datadog_confd_checksd-BEFORE.yaml datadog_confd_checksd-AFTER.yaml
```
<details><summary>resulting diff, all good</summary>

```diff
--- datadog_confd_checksd-BEFORE.yaml   2026-05-29 10:53:59.460619614 +0200
+++ datadog_confd_checksd-AFTER.yaml    2026-05-29 10:49:00.538887458 +0200
 
@@ -370,7 +370,7 @@
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/version: "7"
   annotations:
-    checksum/confd-config: aafb75114f68f607cb8160d786a1fff913ddbdf2ea541461be376c7d627fed1c
+    checksum/confd-config: 5453de9269af856b14ccb866a199cdba3919866dfc38db33b3b1453a109fcfba
     checksum/autoconf-config: 74234e98afe7498fb5daf1f36ac2d78acc339464f950703b8c019892f982b90b
 data:
 
@@ -706,22 +706,22 @@
     init_config:
   
     instances:
-        - package: debian
-          min_collection_interval: 60
-        - package: debian-stable
-          min_collection_interval: 60
-        - package: rpm-stable
-          min_collection_interval: 60
-        - package: rpm
-          min_collection_interval: 60
-        - package: war
-          min_collection_interval: 60
-        - package: war-stable
-          min_collection_interval: 60
-        - package: windows
-          min_collection_interval: 60
-        - package: windows-stable
-          min_collection_interval: 60
+      - package: debian
+        min_collection_interval: 60
+      - package: debian-stable
+        min_collection_interval: 60
+      - package: rpm-stable
+        min_collection_interval: 60
+      - package: rpm
+        min_collection_interval: 60
+      - package: war
+        min_collection_interval: 60
+      - package: war-stable
+        min_collection_interval: 60
+      - package: windows
+        min_collection_interval: 60
+      - package: windows-stable
+        min_collection_interval: 60
[...]

PR #7786, opened 2026-06-02
refactor(buildReportStalenessCheck): reconstruct build report URLs from job path and controller

jayfranco999:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5135

Follow-up to #7778, refactoring `buildReportStalenessCheck` to converge the
config format between kubernetes-management and the datadog repo
(`build-report-jobs.yaml`). This is a prerequisite for auto-syncing jobs
between the two repos via updatecli.

## Changes

- `buildReportStalenessCheck.yaml`: Converted from an `instances:` list to a `build_report_jobs:` map keyed by job name, matching the format in the datadog repo. Each entry now uses `controller` + `job` instead of a full `url`, and `threshold_in_minutes` is renamed to `threshold_minutes`.
- `datadog_confd_checksd.yaml.gotmpl`: The template now parses the map (`fromYaml`) and renders it into the `instances:` list the Datadog Agent expects, via `range`.
- `buildReportStalenessCheck.py`: The report URL is reconstructed at runtimef rom `controller` + `job` instead of reading a hardcoded `url`, and the check now reads `threshold_minutes`.

## Testing

```bash
helmfile template -f ./clusters/publick8s.yaml -l name=datadog > after.yaml
```

### DIFF
--- a/config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.yaml
+++ b/config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.yaml
@@ -3,14 +3,10 @@ init_config:
 instances:
   ## infra.ci.jenkins.io jobs
   - controller: infra.ci.jenkins.io
-    job: docker-jobs/docker-404/main
-    threshold_minutes: 1440
-    # TODO: restore 5 mins after jenkins-infra/helpdesk#2843
-    min_collection_interval: 60
-  - controller: infra.ci.jenkins.io
     job: acceptance-tests/acceptance-tests/check-agent-availability
     threshold_minutes: 1440
     min_collection_interval: 60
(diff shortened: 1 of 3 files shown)


config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.yaml
```diff
@@ -2,70 +2,66 @@ init_config:
 
 instances:
   ## infra.ci.jenkins.io jobs
-  - url: https://builds.reports.jenkins.io/build_status_reports/infra.ci.jenkins.io/docker-jobs/docker-404/main/status.json
```
lemeurherve, line 5: @jayfranco999 the removal of docker-404 should be in its own PR.

PR #7809, opened 2026-06-10
refactor(buildReportStalenessCheck): use controller + job list fields

jayfranco999:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5135
- https://github.com/jenkins-infra/helpdesk/issues/5159

Follow-up to #7778, refactoring `buildReportStalenessCheck` so the job config hares a single canonical list structure with the `jenkins-infra/datadog` repo (`build-report-jobs.yaml`).

## Changes

- `buildReportStalenessCheck.yaml`: Each instance now uses `controller` + `job` instead of a full `url`, and `threshold_in_minutes` is renamed to `threshold_minutes`. The `instances` list structure is retained.
- `buildReportStalenessCheck.py`: The report URL is reconstructed at runtime from `controller` + `job` instead of reading a hardcoded `url`, and the check now reads `threshold_minutes`.
- Dropped the per-instance `min_collection_interval` (Agent reverts to its default collection interval).
- dropped docker-404 from `buildReportStalenessCheck.yaml`

## Companion PR (to be opened)

- jenkins-infra/datadog: mirrors this list structure and adds a `for` expression in `monitor-build-reports.tf` to convert the list into the map that `for_each` requires.

## Testing

```bash
helmfile template -f ./clusters/publick8s.yaml -l name=datadog > after.yaml

### DIFF
--- a/config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.py
+++ b/config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.py
@@ -8,6 +8,7 @@ from urllib.error import URLError, HTTPError
 
 from datadog_checks.base.checks import AgentCheck
 
+BUILD_REPORTS_BASE = 'https://builds.reports.jenkins.io/build_status_reports'
 
 class BuildReportStaleness(AgentCheck):
     """
@@ -39,9 +40,10 @@ class BuildReportStaleness(AgentCheck):
         """
             Datadog custom check
         """
-        url = instance['url']
-        controller = instance.get('controller', 'unknown')
-        threshold_in_minutes = instance['threshold_in_minutes']
+        controller = instance['controller']
+        job = instance['job']
+        url = f"{BUILD_REPORTS_BASE}/{controller}/{job}/status.json"
+        threshold_in_minutes = instance['threshold_minutes']
         base_tags = [f"controller:{controller}"]
 
         self.warning(f"BuildReportStaleness: {controller}")
--- a/config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.py
+++ b/config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.py
@@ -43,7 +43,7 @@ class BuildReportStaleness(AgentCheck):
         controller = instance['controller']
         job = instance['job']
         url = f"{BUILD_REPORTS_BASE}/{controller}/{job}/status.json"
-        threshold_in_minutes = instance['threshold_in_minutes']
+        threshold_minutes = instance['threshold_minutes']
         base_tags = [f"controller:{controller}"]
 
         self.warning(f"BuildReportStaleness: {controller}")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jayfranco999** [comment]
Since 
- https://github.com/jenkins-infra/datadog/pull/367

has been merged let's merge this to proceed with automating the list structure in datadog and also proceed with 
- - https://github.com/jenkins-infra/helpdesk/issues/5159

File: config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.py [hunk 1]
```diff
@@ -39,9 +40,10 @@ def check(self, instance):
         """
             Datadog custom check
         """
-        url = instance['url']
-        controller = instance.get('controller', 'unknown')
-        threshold_in_minutes = instance['threshold_in_minutes']
+        controller = instance['controller']
+        job = instance['job']
+        url = f"{BUILD_REPORTS_BASE}/{controller}/{job}/status.json"
+        threshold_in_minutes = instance['threshold_minutes']
```
lemeurherve:
```suggestion
        threshold_in_minutes = instance['threshold_in_minutes']
```
nit: as this side is the "driver"/source and as this variable naming doesn't have any strict requirement on the datadog repo side, using the expected name directly seems more appropriate.

jayfranco999: good catch, agreed. The source side variable should mirror the YAML key directly.

lemeurherve: @jayfranco999 don't forget to update the YAML file accordingly ;)

config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.yaml
```diff
@@ -1,75 +1,52 @@
...
-    controller: infra.ci.jenkins.io
-    threshold_in_minutes: 1440
-    min_collection_interval: 60
-  - url: https://builds.reports.jenkins.io/build_status_reports/infra.ci.jenkins.io/terraform-jobs/terraform-aws-sponsorship/main/status.json
-    controller: infra.ci.jenkins.io
-    threshold_in_minutes: 1440
-    min_collection_interval: 60
+  - controller: infra.ci.jenkins.io
+    job: acceptance-tests/acceptance-tests/check-agent-availability
+    threshold_minutes: 1440
```
lemeurherve, line 5: Cf https://github.com/jenkins-infra/kubernetes-management/pull/7809#discussion_r3412099478, you've updated this field to `threshold_in_minutes` in the python script but not in this file.

jayfranco999: I've aligned both to `threshold_minutes` rather than `threshold_in_minutes`: the YAML keys are already threshold_minutes (matching the datadog side), so I renamed the Python variable + key read to match that instead of flipping the YAML keys back. Net result is the same, variable and key are now consistent and use the canonical name. Updated the `threshold_in_minutes/60` reference too.

lemeurherve: That's datadog agent here that is waiting for a variable named `threshold_in_minutes`, nothing in the datadog repo is requiring `threshold_minutes`, hence my initial suggestion to use the variable name required by datadog agent.

lemeurherve (approved):

config/publick8s/datadog_confd_checksd/buildReportStalenessCheck.yaml
jayfranco999: As per discussion, I'll open a subsequent PR with the suggested changed, thanks!

PR #8042, opened 2026-09-07
chore(pipeline): lint YAML files from `cluster` and `updatecli` folders too

lemeurherve:
This change ensures all YAML files are linted, not only the ones in `config` folder.


dduportal:
You most probably will end up with this change reverted by updatecli in its PRs, as Yaml linting does not have generally defined rules so each linter and parser has its own behavior. Apporving because I'm assuming it does not kill but be prepared to get noisy PRs (or not if lucky)

dduportal (approved):

PR #8064, opened 2026-09-15
feat(datadog): add a staleness check for the plugin health engines

jayfranco999:
Adds a Datadog custom check for plugin-health-scoring engine staleness. Monitors themselves are a separate PR in jenkins-infra/datadog.

Same shape as `buildReportStalenessCheck` from jenkins-infra/helpdesk#2843: fetch a JSON endpoint, compute the age of a timestamp, submit gauges.

We already have a synthetic asserting `$.status is UP` on `/actuator/health`, and it was green for the whole June to September outage in jenkins-infra/helpdesk#5270. Looking at `EngineHealthIndicator`, `status` comes from the last recorded success and never expires, so an engine that stops being scheduled just keeps saying `UP` with an older and older `lastSuccess`. So we were checking liveness, not freshness. A synthetic can assert on a JSONPath value but can't do arithmetic on a timestamp, hence a check.

Submits `reachable`, `age_in_hours`, `threshold_in_hours`, `stale` and `engine_ok`, tagged per engine. `stale` catches the unscheduled case above. `engine_ok` catches the opposite one, where a run throws or wedges past the 6h `STUCK_RUN_THRESHOLD` and the status goes `DOWN` with a fresh timestamp.

Both engines come back in one response so it's a single instance and a single HTTP call, with the URL as a constant in the check like `buildReportStalenessCheck` does with its base URL. Threshold is 6h to match `STUCK_RUN_THRESHOLD`, which with an hourly `peCron` is six missed runs. `/actuator/prometheus` and `/actuator/metrics` are both 403 so there's nothing to scrape.

Ran it locally against a stubbed `AgentCheck`: live endpoint, a 72h old `lastSuccess` still reporting `UP` (the jenkins-infra/helpdesk#5270 case, sets `stale` and leaves `engine_ok` at 1), endpoint down, `DOWN` with a fresh timestamp, an engine missing from the response, and a malformed timestamp. Last two warn and carry on rather than blowing up the check.

Haven't run it in a real Agent, the harness stubs the import, so I'll fix whatever it does once it's on publick8s.

Ref: jenkins-infra/helpdesk#5257


dduportal (approved):
If it works ;) Nothing shocking but let's iterate quickly on this one

